PR #12, opened 2017-11-01
Fix some README links and kiam-server health check

jbachus:
Small changes to fix the README and allow the health checks on the kiam-server to pass


pingles:
Thanks for the fixes! You're right in that the server changes are out-of-date: to support using gRPC load-balancing we had to change the service to be headless, requiring us to settle on using the service port on the pods; I updated the deployment but forgot the service!

PR #28, opened 2018-01-25
Add optional role for server to assume before assuming pod roles

jpb:
This allows a stable role, independant of controller instance roles, to be used
by the server when calling `sts:AssumeRole`. The `--assume-role-arn` still needs
to have a trust relationship with the instance role where the server is running.


pingles:
Thanks for your PR!

Would you mind adding some stuff into the README, or maybe into docs/IAM.md to explain a little about why you should do this also please- it feels like this is something related to a practice/model that AWS recommend? If it's suggested then it'd be good to highlight people should probably use this.

jpb:
👍  I've added some information as to why you might want to use `--assume-role-arn` to docs/IAM.md.

I don't think this practice would be recommended (or not) by AWS. It doesn't provide any additional security (as the cluster nodes will still need to be able to assume this role, and thus the application roles too, although it means that the cluster nodes can not assume application roles directly). The motivation behind `--assume-role-arn` is to make moving pods between clusters that have different EC2 instance roles easier (by having application roles trust `--assume-role-arn` instead of each of the cluster node EC2 instance roles).

pingles:
Fantastic, makes a lot of sense. Thank you.

Issue #50, opened 2018-04-07
Change Server to only report healthy once a full pod/namespace sync has been performed


pingles:
The cache controller used by both the pod and namespace cache can report whether a sync operation has been performed. Until the sync the server relies on watchers delivering events to record their state. 

This would help control the rollout of server processes (alongside the pod disruption budget) to proceed more safely.

pingles:
Instead of the health check the server boot process was changed to:

1. Start prefetcher
2. Start the pod and namespace caches, additionally these will block until the caches have performed a `Sync` operation
3. Start the gRPC listener

This should ensure that by the time the health check succeeds (because the listener is available) pods and at least some credentials have been retrieved.

pingles:
This was merged in #51.

Issue #48, opened 2018-03-22
Issues with IAM role access after a rolling update of the k8s cluster


dgeorgievski:
I am using kiam v2.6 with kops managed kubernetes 1.9.5 cluster in AWS. Whenever a rolling update is performed `kops rolling-update` some pods have issues accessing their respective IAM roles.

From the agent error log messages it seems that the mapping between the IAM roles and pods is lost. The affected pod is trying to access one IAM role, but Kiam believes it is annotated with another IAM role.
 
```
kiam-agent-mhtfd kiam {"addr":"100.124.0.4:36284","level":"error","method":"GET","msg":"error processing request: assume role forbidden: requested 'iam-role-one' but annotated with 'iam-role-two', forbidden","path":"/latest/meta-data/iam/security-credentials/iam-role-one","status":403,"time":"2018-03-22T19:45:29Z"}
```

The workaround is to restart all kiam servers after which the correct mapping is restored. I was able to replicate the issue twice in a row. 

The server is started with the following settings:
```
        - --json-log
        - --level=info
        - --bind=0.0.0.0:443
        - --cert=/etc/kiam/tls/server.pem
        - --key=/etc/kiam/tls/server-key.pem
        - --ca=/etc/kiam/tls/ca.pem
        - --role-base-arn-autodetect
        - --sync=1m
        - --prometheus-listen-addr=0.0.0.0:9620
        - --prometheus-sync-interval=5s
```

while the agent with:
```
        - --level=info
        - --iptables
        - --host-interface=weave+
        - --json-log
        - --port=8181
        - --cert=/etc/kiam/tls/agent.pem
        - --key=/etc/kiam/tls/agent-key.pem
        - --ca=/etc/kiam/tls/ca.pem
        - --server-address=kiam-server:443
        - --prometheus-listen-addr=0.0.0.0:9620
        - --prometheus-sync-interval=5s
```

Please let me know if you need any other information.

Regards,

Dimitar

pingles:
Interesting. This sounds similar to #47: pods are starting before the watch deltas have been delivered- so the pod cache has the wrong data. Restarting the servers would obviously clear everything and cause a fresh sync fixing things.

Could you re-run and prove that it works after 1 minute (the `--sync` flag of server controls how frequently it runs a full sync of pods) or reduce the sync interval down.

I'll try and investigate more tomorrow and see what I can find.

[...]

PR #51, opened 2018-04-17
Use IndexerInformer rather than controller and queue

pingles:
This PR makes a few changes:

* The implementation of PodCache and NamespaceCache now use `cache.NewIndexerInformer`. This removes a bunch of the boilerplate processing code that connected the controller to a store (and indexer). It addresses an error reported in #46 (type conversion inside the old process function), and a similar problem reported in #48.

* Changes the PodCache and NamespaceCache `Run` funcs to only return when they have been synced. The Server has changed its startup flow to start the prefetcher routines before loading the caches. The gRPC service will only start listening for clients after the cache's have synced, potentially avoiding a race where the pods receive requests before they've had a chance to initialise properly. This achieves the same as noted in #50.


pingles:
Yeah, agreed. 

Funnily enough I actually started more general but it turned out to be more work than I thought it was worth- the pod handler behaves slightly differently (both logging and then forwarding pods for the prefetcher). In the future it's probably worth refactoring quite a lot of the code but this keeps it obvious for now.

PR #54, opened 2018-04-26
Supports use of ! prefix for interface name

polarbizzle:
There are some CNI implementations that make use of multiple
Elastic Network Interfaces and secondary IPs to assign pods ips
directly from the VPC ranges. For example:

    https://github.com/aws/amazon-vpc-cni-k8s
    https://github.com/lyft/cni-ipvlan-vpc-k8s

For kiam to function correctly, it is necessary to have an iptables
rule that applies to all of the interfaces that pod traffic may come
from. And since these interfaces may be added and removed on demand,
it is necessary to have DNAT rules that will continue to work as
interfaces come and go.

iptables supports inverted matching for interface names which can be
useful to include all but certain interfaces in rules. For example:

    iptables --append PREROUTING --protocol tcp \
      --destination 169.254.169.254 --dport 80  \
      \! -i loopback   --jump DNAT  --table nat \
      --to-destination 10.100.100.3:8181

will apply the DNAT rule to all interfaces except the loopback.

This change puts the "!" for inverting the interface before the name
of the interface in the rules spec that inverted rules work as
intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,15 @@ Please also make note of how to configure IAM in your AWS account; notes in [doc
 Kiam is split into two processes that run independently.
 
 ### Agent
-This is the process that would typically be deployed as a DaemonSet to ensure that Pods have no access to the AWS Metadata API. Instead, the agent runs an HTTP proxy which intercepts credentials requests and passes on anything else. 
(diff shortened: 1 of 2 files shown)


pingles:
Thanks! Would you mind updating the README also to add include section on the iptables config wrt those cni plugins, it'd be really useful to call out what you could/should use to make them work. I'll add notes on what we use with Flannel/Canal also.

polarbizzle:
Let me know what you think about that documentation blurb. We've only used weave and cni-ipvlan-vpc-k8s so I added entries for those two. Thanks for the feedback 👍

pingles:
That's perfect, thank you.

PR #55, opened 2018-04-26
adds support for default roles

polarbizzle:
Addresses #1 
If a pod doesn't have a role specified, it can be useful in some
cases to return a default role instead of empty role. This change
adds two flags to the agent:

   --default-role lets you specify the name of the role
   --default-role-autodetect if you haven't specified a default-role
       this flag will cause kiam to attempt to look up the role using
       the instance profile arn from the metadata service

For pods that get the default role, the real ec2 metadata service is
used to lookup credentials.

### DIFF
--- a/pkg/aws/metadata/handler_credentials.go
+++ b/pkg/aws/metadata/handler_credentials.go
@@ -33,12 +33,18 @@ type credentialsHandler struct {
 	credentialsProvider sts.CredentialsProvider
 	clientIP            clientIPFunc
 	policy              server.AssumeRolePolicy
+	defaultRole         string
 }
 
 func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter, req *http.Request) (int, error) {
 	credentialTimings := metrics.GetOrRegisterTimer("credentialsHandler", metrics.DefaultRegistry)
 	startTime := time.Now()
 	defer credentialTimings.UpdateSince(startTime)
+
+	requestedRole := mux.Vars(req)["role"]
+	if requestedRole == "" {
+		return http.StatusBadRequest, fmt.Errorf("no role specified")
+	}
 
 	err := req.ParseForm()
 	if err != nil {
@@ -56,14 +62,9 @@ func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter, 
 		return http.StatusInternalServerError, fmt.Errorf("error finding pod for ip %s: %s", ip, err.Error())
 	}
 
-	if foundRole == "" {
+	if foundRole == "" && c.defaultRole != requestedRole {
 		metrics.GetOrRegisterMeter("credentialsHandler.emptyRole", metrics.DefaultRegistry).Mark(1)
 		return http.StatusNotFound, EmptyRoleError
-	}
-
-	requestedRole := mux.Vars(req)["role"]
-	if requestedRole == "" {
-		return http.StatusBadRequest, fmt.Errorf("no role specified")
 	}
 
 	decision, err := c.policy.IsAllowedAssumeRole(ctx, requestedRole, ip)
--- a/pkg/aws/metadata/handler_credentials.go
+++ b/pkg/aws/metadata/handler_credentials.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"github.com/gorilla/mux"
 	"github.com/rcrowley/go-metrics"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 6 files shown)


mikesplain:
Nice, this will be useful for us too. Thanks @polarbizzle!

sibrahim001982:
YAY! I'm all about new features!

But, what is the use case? I mean, what if a pod does not have a role? what happens?

polarbizzle:
If a pod doesn't have a role annotation then it doesn't get any permissions at all. Which eventually for us will be ideal. In our case its going to allow phased adoption from our current setup where all the permissions are dumped into one role. We need pods that we haven't been able to split out into their own role to still work in prod during the unfortunately somewhat lengthy process of migration.

sibrahim001982:
Thanks for the reply my friend @polarbizzle 

I swear to God, I didn't understand a word :)

mikesplain:
Yeah we use kube2iam currently and use this type of functionality for our master nodes, that way we don’t have to modify the master api specs that are setup by kops and allow them to fall back to the IAM role of the node.

pingles:
Thanks for contributing.

pingles:
Sorry- posted a little too soon :) I have a few comments on the code

pkg/aws/metadata/handler_credentials.go
```diff
@@ -56,12 +58,29 @@ func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter,
 		return http.StatusInternalServerError, fmt.Errorf("error finding pod for ip %s: %s", ip, err.Error())
 	}
 
+	requestedRole := mux.Vars(req)["role"]
+
+	if foundRole == "" && c.defaultRole == requestedRole {
```
pingles: Is there a reason for this block to interact directly with the metadata API rather than the server? 

If possible I'd rather keep the model simpler and ensure all credential accesses go via the server- so the agent could still determine a default role, but the credentials would still be requested from the server.

pkg/aws/metadata/handler_credentials.go
```diff
@@ -56,12 +58,29 @@ func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter,
...
+		metrics.GetOrRegisterMeter("credentialsHandler.defaultRole", metrics.DefaultRegistry).Mark(1)
+		resp, err := http.Get(fmt.Sprintf("http://169.254.169.254/latest/meta-data/iam/security-credentials/%s", c.defaultRole))
+		if err != nil {
+			return resp.StatusCode, err
+		}
+		body, err := ioutil.ReadAll(resp.Body)
```
pingles: I'd prefer not to have any of the http.Get code but if it was necessary I think it's missing a `defer resp.Body.Close()` before the reader attempts to read.

pkg/aws/metadata/handler_credentials.go
pingles: I added this comment to the start of the block, the code I was referring to is this:

```
resp, err := http.Get(fmt.Sprintf("http://169.254.169.254/latest/meta-data/iam/security-credentials/%s", c.defaultRole))
```

polarbizzle:
Thanks @pingles for the feedback. I've done some tweaking and inline with your comments and I agree it's much simpler to just send the request on to the server

pingles:
@polarbizzle fantastic, thank you

polarbizzle:
@pingles do you think this can make it into a release in the near future?

pingles:
Hi!

I’ll try and take a look this week. Thanks for the reminder.

polarbizzle:
When I tested this before I realized today that I hadn't actually deployed the updated docker image like I thought. When the default role request is sent to the server it fails the `RequestingAnnotatedRolePolicy` unless you redeploy pods with the annotation set to the default role. Our intent was to make it so you wouldn't have to redeploy pods to get the default role so I'm revisiting this. I suppose it can just be a flag passed to the server. That seems better than passing the default role along with the credentials requests

pingles:
Well I guess it's a question of whether agents would have different default roles, or whether the server would enforce a single default everywhere. I guess the former is more flexible but the latter feels simpler (so I'd err towards that).

polarbizzle:
The other consideration I failed to mention is that you have to modify the roles for the nodes to allow the master role to assume them. 

Now I'm leaning back towards the original solution. That way nothing has to be changed on the server, existing IAM roles don't need to be modified, and pods don't need to be redeployed and you have the flexibility you mentioned. 

I just did a test again (for sure this time) and this does this trick.

pingles:
> you have to modify the roles for the nodes to allow the master role to assume them.

Modifying the roles to specify that either the node running the server process, or passing `--assume-role-arn` feels ok though?

What do you mean by the original solution? :)

polarbizzle:
Sorry about the lack of clarity in my last comment. I added a second commit that has the code that I had originally submitted. It uses the real metadata service on the agent nodes to detect the default role.

We use kops so our masters (where we run the kiam servers) have a masters role and the nodes have a node role. The node role today has all the permissions any pod needs plus whatever kops needs. I was referring to having to modify the node role trust policy to allow masters to assume it.  This may not be the best reasoning to consider for this feature.

The main reason that is making me prefer using the real metadata service on the agents is that it doesn't require changes to the server. If I wanted to depend on the result from the server, there would need to be a way to let the server know which role is the default. Perhaps `--assume-role-arn` could be used, but that would overload the purpose of that flag and potentially be confusing. In any case, after the server knew what the default role was, it would still require modifications to the RequestingAnnotatedRolePolicy to allow pods with no annotation to get the default role.  It feels like more moving parts. Using the real metadata service on the agents for the default role keeps all the logic about that in the agent.

pingles:
I think I'd still prefer a solution that:

1. IAM roles in AWS would be configured to permit either the nodes running Kiam server to assume them, or, the role specified by `--assume-role-arn` (which the server would assume).
2. Specify a default role via a server flag
3. Change policy which permits pods to assume the default role if they're not annotated with anything else.

I'd prefer this so that all credentials still flow via the server process and ensures that the nodes you run the kiam agent on (where your user workloads are) only need the base IAM permissions needed for Kubelet.

Joseph-Irving:
closing due to inactivity

PR #76, opened 2018-06-01
wait for balancer to have some address available on gateway

sp-joseluis-ledesma:
When running the health command, more often than not, the output is:
```
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no address available 
INFO[0000] healthy: ok
```
running with GRPC_GO_LOG_SEVERITY_LEVEL=info GRPC_GO_LOG_VERBOSITY_LEVEL=8 the problem is that the balancer has not been notifiied of the address before it tries to connect:
``` 
INFO: 2018/06/01 06:55:06 ccBalancerWrapper: updating state and picker called by balancer: IDLE, 0xc420205320
INFO: 2018/06/01 06:55:06 dialing to target with scheme: ""
INFO: 2018/06/01 06:55:06 could not get resolver for scheme: ""
WARN[0000] error checking health: rpc error: code = Unavailable desc = there is no address available 
INFO: 2018/06/01 06:55:06 balancerWrapper: is pickfirst: false
INFO: 2018/06/01 06:55:06 grpc: failed dns SRV record lookup due to lookup _grpclb._tcp.localhost on 100.64.0.10:53: no such host.
[...]
```

This PR adds a "wait" for the balancer to have address on the NewGateway function, so when it returns the KiamGateway we ensure there is some address in the balancer

### DIFF
--- a/cmd/agent/main.go
+++ b/cmd/agent/main.go
@@ -42,6 +42,7 @@ type options struct {
 	hostInterface        string
 	serverAddress        string
 	serverAddressRefresh time.Duration
+	timeoutKiamGateway   time.Duration
 	prometheusListen     string
 	prometheusSync       time.Duration
 
@@ -85,6 +86,7 @@ func main() {
 
 	kingpin.Flag("server-address", "gRPC address to Kiam server service").Default("localhost:9610").StringVar(&opts.serverAddress)
 	kingpin.Flag("server-address-refresh", "Interval to refresh server service endpoints").Default("10s").DurationVar(&opts.serverAddressRefresh)
+	kingpin.Flag("gateway-timeout-creation", "Timeout to create the kiam gateway ").Default("50ms").DurationVar(&opts.timeoutKiamGateway)
 	kingpin.Flag("cert", "Agent certificate path").Required().ExistingFileVar(&opts.certificatePath)
 	kingpin.Flag("key", "Agent key path").Required().ExistingFileVar(&opts.keyPath)
 	kingpin.Flag("ca", "CA certificate path").Required().ExistingFileVar(&opts.caPath)
@@ -129,7 +131,10 @@ func main() {
 	config := http.NewConfig(opts.port)
 	config.AllowIPQuery = opts.allowIPQuery
 
-	gateway, err := kiamserver.NewGateway(opts.serverAddress, opts.serverAddressRefresh, opts.caPath, opts.certificatePath, opts.keyPath)
+	ctxGateway, cancelCtxGateway := context.WithTimeout(context.Background(), opts.timeoutKiamGateway)
+	defer cancelCtxGateway()
+
+	gateway, err := kiamserver.NewGateway(ctxGateway, opts.serverAddress, opts.serverAddressRefresh, opts.caPath, opts.certificatePath, opts.keyPath)
 	if err != nil {
 		log.Fatalf("error creating server gateway: %s", err.Error())
 	}
--- a/cmd/health/main.go
+++ b/cmd/health/main.go
@@ -27,6 +27,7 @@ type options struct {
 	logLevel             string
 	serverAddress        string
 	serverAddressRefresh time.Duration
+	timeoutKiamGateway   time.Duration
 	certificatePath      string
 	keyPath              string
 	caPath               string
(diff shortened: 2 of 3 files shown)


sp-joseluis-ledesma:
I added a parameter to the NewGateway function, and passed it from the agent and health commands. Is it ok? or do you prefer another aproach?

Perhaps the NewGateway parameters could be converted to the "WithOption" pattern, but I leave this for another PR :)

pkg/server/gateway.go
```diff
@@ -45,7 +45,7 @@ type GatewayConfig struct {
 	TLS     *TLSConfig
 }
 
-func NewGateway(address string, refresh time.Duration, caFile, certificateFile, keyFile string) (*KiamGateway, error) {
+func NewGateway(address string, refresh time.Duration, waitForBalancer time.Duration, caFile, certificateFile, keyFile string) (*KiamGateway, error) {
```
pingles: Could we pass the `context.Context` through as the first arg to `NewGateway`, rather than the wait duration.

pkg/server/gateway.go
```diff
@@ -80,6 +80,13 @@ func NewGateway(address string, refresh time.Duration, caFile, certificateFile,
 		return nil, err
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), waitForBalancer)
+	defer cancel()
+	_, _, err = balancer.Get(ctx, grpc.BalancerGetOptions{BlockingWait: true})
```
pingles: If we pass the context through then this doesn't need to create `context.WithTimeout` it can just cancel according to the context it's provided.

cmd/agent/main.go
```diff
@@ -129,7 +129,7 @@ func main() {
 	config := http.NewConfig(opts.port)
 	config.AllowIPQuery = opts.allowIPQuery
 
-	gateway, err := kiamserver.NewGateway(opts.serverAddress, opts.serverAddressRefresh, opts.caPath, opts.certificatePath, opts.keyPath)
+	gateway, err := kiamserver.NewGateway(opts.serverAddress, opts.serverAddressRefresh, 50*time.Millisecond, opts.caPath, opts.certificatePath, opts.keyPath)
```
pingles: Instead of passing the time we'll pass a `context.Context` through. We should also add a CLI flag to control what the wait duration is.

cmd/health/main.go
```diff
@@ -63,7 +63,7 @@ func main() {
 		log.SetLevel(log.ErrorLevel)
 	}
 
-	gateway, err := kiamserver.NewGateway(opts.serverAddress, opts.serverAddressRefresh, opts.caPath, opts.certificatePath, opts.keyPath)
+	gateway, err := kiamserver.NewGateway(opts.serverAddress, opts.serverAddressRefresh, 50*time.Millisecond, opts.caPath, opts.certificatePath, opts.keyPath)
```
pingles: As with the agent's main, we should construct a context and pass it to `NewGateway` and set the timeout on that context from a cli flag.

sp-joseluis-ledesma:
Ok, updated following your comments. Let me know if there is anything more to change.

mikesplain:
Thanks @sp-joseluis-ledesma! We run into this as well, this is great.

pingles:
Thanks @sp-joseluis-ledesma

pingles:
I've tagged it as `v2.8` and the image is now up on quay.io, I'll create a separate PR with the CHANGELOG updates but you can update now if you like!

TattiQ:
Hi guys, sorry for bothering you in this thread but my issues seems a bit connected to this one. I am using version 2.7 now with kops , it ran fine in dev and today I tried it out in prod (hoping to substitute kube2iam) and got an issue with some of the agents when they can not seem to connect to the server. The error suggests that there are no addresses to connect to and unlike for Jose Luis, I get a fatal error. and this is reproduced constantly on those agents even if the ping succeeds after all.  I wonder if anyone  has  any thoughts on that. 
screen shot 2018-08-03 at 10 33 09

pingles:
@TattiQ only thing I can recommend for now is increasing the time limits on your health check and see if that helps- do liveness probes always fail or intermittently? Does it do the same if you specify the FQDN for the server?

Could you perhaps open as a new issue and then we can track it?

Issue #44, opened 2018-03-09
Allow configurable cache TTL


pingles:
According to the [AWS documentation AssumeRole](https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html) has a minimum session duration of 15 minutes. Kiam's server currently expires credentials from it's cache after 10 minutes- if a pod is still running that needs the credentials they'll be updated ahead of the client needing them.

#41 introduced the ability to generate credentials with customised durations (longer than the 15 minute default) but the AWS API is still called every 10 minutes to request credentials.

For users with large numbers of roles/lots of AWS calls it'd be nice to change Kiam so that it expires credentials from the cache relative to their session duration: if the session expires in an hour the cache should evict the credentials at 55 minutes.

As an extension, we could support an additional annotation that would allow users to specify the session duration (#38 also mentions adding session names).

pingles:
This has been merged so closing now.

PR #77, opened 2018-06-01
setup cache TTL to be 5 minutes before the session expires

sp-joseluis-ledesma:
This will set the cacheTTL to sessionDuration - 5 minutes. For the default values Kiam will have the very same behavior, but right now if you increase the sessionDuration the cache is expired after 10 minutes anyway. Solves #44

### DIFF
--- a/cmd/server/main.go
+++ b/cmd/server/main.go
@@ -24,6 +24,7 @@ import (
 	"github.com/pubnub/go-metrics-statsd"
 	"github.com/rcrowley/go-metrics"
 	log "github.com/sirupsen/logrus"
+	"github.com/uswitch/kiam/pkg/aws/sts"
 	"github.com/uswitch/kiam/pkg/prometheus"
 	serv "github.com/uswitch/kiam/pkg/server"
 	"gopkg.in/alecthomas/kingpin.v2"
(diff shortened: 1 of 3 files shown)


sp-joseluis-ledesma:
I don't know if you may prefer a configurable parameter, but I thought it may make sense to just use sessionDuration - 5 minutes

pingles:
Thanks for more contributions :)

I can't remember whether Kiam checks this already but it would be good to make sure that the session duration value is validated to be at least 15 minutes. Likewise, the default of `duration - 5*time.Minute` feels reasonable.

pingles:
So we don't currently validate but it would be nice if we did :)

Would you mind adding something to the server's main to validate the `session-duration` arg. We can probably just validate the duration is at least 15 minutes inside the server's main, logging a fatal if it's too short.

pingles:
Actually- we should probably place that const into the `sts` package and reference from the server cmd, so it'd be something like `sts.AWSMinSessionDuration`?

sp-joseluis-ledesma:
It makes sense. PR updated

pingles:
Thanks @sp-joseluis-ledesma

Issue #75, opened 2018-05-30
Leading / from IAM path has to be omitted in annotations


2rs2ts:
Good: `k8s-roles/myserver`
Bad: `/k8s-roles/myserver`

My company uses the IAM path to restrict which roles the nodes running kiam-server can assume. One of the most common mistakes developers (including myself!) make is forgetting that you must omit the leading `/` from the path in the annotations.

I think this should either be documented or merely accepted either way. I'm happy to implement whatever acceptance criteria you come up with, when I have the time to. I would have just submitted a PR but I wanted feedback and AC first.

roffe:
Imho we should accept leading slash and just strip it, I made a PR that's running in my staging cluster without issues so far.

Image is at `roffe/kiam:issue-75-1` for anyone that wants to give it a spin

pingles:
Interesting.

I guess if `/` is reserved in role names then it'd be safe to have Kiam detect and remove the prefix if it exists. Full ARN names can also be used to be explicit.

1. `k8s-roles/myserver` would expand to `arn:aws:iam::account-id:role/k8s-roles/myserver`
2. `/k8s-roles/myserver` should expand to `arn:aws:iam::account-id:role/k8s-roles/myserver`
3. `arn:aws:iam::account-id:role/k8s-roles/myserver` would expand to itself.

Is that correct? Is there ever a time where 2 should _not_ remove the prefix?

roffe:
Afaik there is no situation where the leading slash should not be removed

The problem i think is that DefaultResolver would return a full base ARN ending in a / and when we append a role starting with / we would end up with something like:

`/k8s-roles/myserver expand to arn:aws:iam::account-id:role//k8s-roles/myserver`

and here the aws api would complain about the ARN containing double slashes

PR #80, opened 2018-06-04
handle pod roles starting with /

roffe:
Should be 100% backwards compatible. builded and tested in my staging cluster and it accepts pod roles with leading / and does what i expect it to :)

Closes https://github.com/uswitch/kiam/issues/75

### DIFF
--- a/pkg/aws/sts/arn_resolver.go
+++ b/pkg/aws/sts/arn_resolver.go
@@ -31,6 +31,10 @@ func DefaultResolver(prefix string) *Resolver {
 
 // Resolve converts from a role string into the absolute role arn.
 func (r *Resolver) Resolve(ctx context.Context, role string) (string, error) {
+	if strings.HasPrefix(role, "/") {
+		role = strings.TrimPrefix(role, "/")
+	}
+
 	if strings.HasPrefix(role, "arn:") {
 		return role, nil
 	}
--- a/pkg/k8s/pod_cache.go
+++ b/pkg/k8s/pod_cache.go
@@ -16,7 +16,6 @@ package k8s
 import (
 	"context"
 	"fmt"
-	"strings"
 	"time"
 
 	metrics "github.com/rcrowley/go-metrics"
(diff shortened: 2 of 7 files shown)


pingles:
@roffe thanks for adding this- could you add some tests to `test/unit/arn_resolver_test.go` to demonstrate the new behaviour pleaes?

roffe:
There we go, Now with 3 tests added

Test 1: Add a role that begins with a /
Test 2: Add a role with a path not beginning with slash
Test 3: Add a role with a path beggining with slash

Hope this satisfies what you asked for @pingles

roffe:
ok, need to put some more work in, after i moved it from the pod_cache to the arn_resolver i'm getting this error when running in my staging cluster

```
{"addr":"100.96.9.14:39726","level":"error","method":"GET","msg":"error processing request: assume role forbidden: requested 'staging.sgn.io_seacucumber-serve' but annotated with '/staging.sgn.io_seacucumber-serve', forbidden","path":"/latest/meta-data/iam/security-credentials/staging.sgn.io_seacucumber-serve","status":403,"time":"2018-06-04T16:44:36Z"}
```

pingles:
ah- there’s policy which checks whether regexes permit etc. There’s also
one that the annotation matches what was requested- it’s probably necessary
to change that to use the resolver, so both sides use the resolved value.

roffe:
built `roffe/kiam:issue-75-4` for the last commit b03c3f8 and i'm running it in my staging cluster now

Added test cases in all combinations i could think of and made so the role is passed through the resolver for sanitazion in policy.go

I hope this is at least is close to what you had envisioned :)

pingles:
Thanks @roffe. I've taken a look through- given the policies are now using the resolvers, perhaps it'd be cleaner/safer for the policies to be constructed with a resolver and then the server can pass it's configured resolver. I feel like without that it's possible for the annotated policy to differ from what the resolver interprets?

Other than that it looks good (and I appreciate you adding more tests 😄 )

roffe:
yeah that makes sense, i was taking some shortcuts with `ARNresolver := sts.DefaultResolver("")` as i wasn't sure about how you felt about doing said incision. I'll see what i can do! :)

pingles:
Thanks @roffe. If not I'll see if I can pick up from where you've got to.

roffe:
i removed the part i added from NamespacePermittedRoleNamePolicy since it's regex matching. Using the configured resolver makes no sense, nor does it to actually strip anything here, this is on the users end to configure their regex properly allow the name they actually put in the role.

I kept the tests / added for it and updated the test regext to allow the role with or without a leading slash

roffe:
First impression is it runs fine in my staging cluster so far :)

pingles:
Changes look good to me. It highlights that `Resolve` never returns an error so I'll refactor that interface tomorrow to tidy things up before I merge.

Thanks @roffe

Issue #27, opened 2018-01-24
Security Context on the Agent


chrislovecnm:
I am noticing

```yaml
 securityContext:
            privileged: true
```

Could the container be run without privileged if it did not set up the DNAT rule with iptables?  Or is there another role that can be used that has less root-like permissions?

pingles:
Yep- that would only be used to configure the DNAT rule. I'm not sure about how it'd work with a different role- do you have something that'd help me figure out how to do it? I'd be glad to sort it.

Alternatively, I also thought about an alternate model where all pods were injected with sidecars- they could use iptables (without needing privileged access) and disable outbound metadata api access without doing it on the nodes directly.

The sidecar stuff could potentially go further and run a whole agent as a sidecar but that's probably a topic for another time :)

pingles:
I was thinking about this over the weekend- another option is potentially integrating via ECS' integration: `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` and https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/index.html?com/amazonaws/auth/ContainerCredentialsProvider.html.

I've not tried it but if I get a chance this week I might give it a whirl!

emmanuel:
I like the idea of removing the requirement for `privileged: true` from the agent, but I don't think that using the ECS credentials distribution model will escape it. It seems that the ECS agent still uses an iptables rule in the host net namespace to redirect a link-local IP (similar to but distinct from the default metadata service link-local ip `169.254.169.254`):
https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html#enable_task_iam_roles
Specifically this part:
```
sudo sysctl -w net.ipv4.conf.all.route_localnet=1
sudo iptables -t nat -A PREROUTING -p tcp -d 169.254.170.2 --dport 80 -j DNAT --to-destination 127.0.0.1:51679
sudo iptables -t nat -A OUTPUT -d 169.254.170.2 -p tcp -m tcp --dport 80 -j REDIRECT --to-ports 51679
```
I hope I'm wrong!

pingles:
@emmanuel looks like you're right, that ECS environment variable doesn't set the whole credentials- just the path:

https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-core/src/main/java/com/amazonaws/auth/ContainerCredentialsProvider.java#L94

roffe:
Changing the security context from `privileged: true` on to 

```
[...]

PR #82, opened 2018-06-04
limit the agents privileges

roffe:
For https://github.com/uswitch/kiam/issues/27

### DIFF
--- a/deploy/agent.yaml
+++ b/deploy/agent.yaml
@@ -30,7 +30,8 @@ spec:
       containers:
         - name: kiam
           securityContext:
-            privileged: true
+            capabilities:
+              add: ["NET_ADMIN"]
           image: quay.io/uswitch/kiam:v2.8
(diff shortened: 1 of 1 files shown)


pingles:
Awesome, thanks @roffe! I just want to confirm this also works with our production clusters before I merge.

2rs2ts:
FYI, this doesn't work for us, we get errors like `{"level":"fatal","msg":"error configuring iptables:running [/sbin/iptables -t nat -C PREROUTING -p tcp -d 169.254.169.254 --dport 80 -j DNAT --to-destination 10.42.25.81:8181 -i cali+ --wait]: exit status 3: iptables v1.6.2: can't initialize iptables table `nat': Permission denied (you must be root)\nPerhaps iptables or your kernel needs to be upgraded.\n","time":"2019-02-22T22:29:59Z"}`

PR #84, opened 2018-06-05
Improve future impl

ash2k:


pingles:
Thanks. There's already some tests for future in `test/unit/future_test.go`, I don't see any scenarios in yours that aren't already covered, mind if you/I remove the ones you've added?

ash2k:
I heven't seen those tests - typically unit tests in Go are in the same package/directory so I was not looking for them anywhere else. Removed!

pingles:
:) yeah- i think i'll probably move them back and then leave the non-package tests in the root or something. more confusing than tidy!

PR #88, opened 2018-06-08
Fix balancer

pingles:
* `grpc.BalancerGetOptions{BlockingWait: true}` appeared to have no behaviour at all- causing the gateway to immediately return. This has been replaced with a retry loop that only returns when an address can be successfully resolved.

* Switch to managing dependencies with `dep`
  * As part of this most libraries have also been upgraded- notably the Kubernetes client libs have been pinned to releases rather than master.


mikesplain:
@pingles Thanks for this fix, we've been running this for about 18 hours now and seen significant performance improvements 👍

pingles:
@mikesplain ah nice- do you have anything you could share? Do you mean the healthcheck stuff has improved or that other performance stuff has improved from the version upgrades?

mikesplain:
@pingles Ahh yes sorry, I should've been more clear. Our most common issue we've seen since switching to Kiam is random `The security token included in the request is invalid` errors when our apps access services like SQS.  We haven't gotten that since I deployed this change.  `kiam_server_aws_assumerole` feels a little slower compared to before this but it's only slightly noticeable in our graphs and I can't even break the number out in a meaningful way. So far, win win for us.

PR #99, opened 2018-06-27
Merge commands into a single binary

### LINKED ISSUE #98: Merge commands
This will help reduce the size of the published Docker image.

There are 3 separate binaries that get built:

* server
* health
* agent

The docker image contains the three binaries which will mostly be copies of exactly the same code.

Rather than having separate binaries we should join them together and make them subcommands.

sambooo:
This set of changes merges `agent`/`server`/`health` into a single `kiam` binary and updates references to the old commands.

Resolves #98.


pingles:
Thanks to @sambooo for a bumper set of changes ;)

Issue #14, opened 2017-11-16
Record IAM errors as events


pingles:
Although the Kiam server will currently record when its unable to retrieve credentials for a role (perhaps because the role doesn't exist, or isn't trusted by the node trying to `AssumeRole`) it means accessing Kiam logs.

It would be lovely to be able to see these errors reported as events for the Pod so that end-users can more easily track down issues.

pingles:
I'm going to write my initial thoughts in case anyone wandering past reads and can help nudge my thinking along.

At the moment most interesting operations on behalf of Pods (determining the role, requesting credentials) are performed through exchanging role names, pod IPs etc. rather than a pod reference.

It feels like there are a couple of ways to change Kiam to support something like the event recording:

1) Agent reports errors when it is unable to respond successfully in one of it's metadata api impersonating handlers. This would probably need the Server changing its operations to include a `ReportEvent` operation, passing an IP to identify the pod.

2) Change the Server interface so that its broader and mirrors the metadata operations (e.g. GetRole, GetCredentials). The Agent HTTP interface would reduce to just calling that method and using the returned error to determine what happened. The Server could then store a reference to the Pod for other operations, recording events etc. I feel like this would simplify things down quite a lot, and could probably be added alongside the existing (although makes the existing methods somewhat deprecated/redundant).

In short- 1) maintains the current chatty form of rpc - GetRole, GetCredentials, IsAllowed etc., 2) reduces that to an operation per handler. 1 involves fewest changes to what's there, 2 is more drastic.

Having written it out I feel like 2 is probably the right thing to do.

emmanuel:
This article from the Kubernetes blog may be useful. It contains some good explanation and code samples about using events to communicate from cluster components (e.g., kiam) to pods: 
https://kubernetes.io/blog/2018/01/reporting-errors-using-kubernetes-events/

pingles:
This is still really important to support- we occasionally have teams that get confused why their applications don't work and its because there are typos in role names etc.

I think this change to the server would help tidy a load of stuff up to. It'll mean adding a new set of operations to the server to eventually deprecate the current.

pingles:
With #109 this gets closer for most operations. @sambooo be good to go through what you and @DewaldV started to update so all stuff is tracked with the event recorder and get this closed!

pingles:
The servers will now also need additional RBAC policy that lets them modify events. We should make sure this is documented in the #117 notes as well as in the current deployment manifests.

[...]

Issue #35, opened 2018-02-06
Replace metrics with native Prometheus


pingles:
Given Prometheus as the standard means for collecting custom metrics within Kubernetes clusters, and the issues around exporting the go-metrics ones, it'd be prudent to overhaul and replace them with native prom equivalents.

I don't think we'd aim for naming parity so probably sensible to just replace according to the naming scheme the Prometheus authors recommend and fold into a major version release.

idiamond-stripe:
If possible, can we continue to have the option to provide statsd metrics, at least for latency metrics.

It's useful for us to ship these to our upstream metrics provider and do aggregation there, rather than in the application, like either prometheus or `go-metrics` do. I have a branch that adds native Prometheus metrics and individual statsd metrics for latency. Would it be useful to open a PR for it?

pingles:
@idiamond-stripe yep, that's ok. Sticking to go-metrics with StatsD sounds ok then?

I'd definitely still like to have native Prometheus metrics, and I think I'd probably rename a bunch of them/change what gets captured to be a little more useful.

I'd love a PR to see what you've done so far (and take on to close this :)

pingles:
#114 should help replace a lot of the gRPC client/server boilerplate.

I'd suggest therefore that we integrate #114 and then the focus of the changes in this issue would be around native prometheus metrics for the HTTP handler. Again, maybe there's a way we can just use the built in integrations to provide telemetry rather than recording/reporting within each of the handlers.

The final thing we'd need to instrument would be timings around the `AssumeRole` operation call.

pingles:
This is addressed in #131- once that's merged we should have equivalent metrics captured.

pingles:
#131 has been merged!

PR #102, opened 2018-06-29
Record KiamRoleFound pod events

sambooo:
This is a prerequisite for #14


pingles:
I've also started on changing the gRPC API to make it easier to expand this to solve #14. Changes incoming, although I've temporarily removed some of the telemetry stuff- it makes it easier to change and we'll be doing #35 anyway.

Issue #91, opened 2018-06-13
Enable Service Account Tokens to Validate Agent


moofish32:
The system currently has mutual TLS for client server authentication. I would like to add the ability for the server to verify the agent via a service account token. This would currently be in addition to the mutual TLS. After we have this working we could enable options to disable mutual TLS for standard TLS. The initial implementation would make this optional via configuration. The changes required would be: 

1. Update protos to support the token field
1. Update server to use the token reviewer api if configured; likely borrow [this code from vault](https://github.com/hashicorp/vault-plugin-auth-kubernetes/blob/master/token_review.go)
1. Update agent to always send service account token if it's present in the pod
1. Update docs for how to enable the feature

Is this a feature others would use?

pingles:
I think this could be done alongside a wider proto rejig (to make it easier to accomplish some other things). I'll try and write up that in another issue to get some feedback also.

Definite +1 from me though.

jacksontj:
Is this still being worked on? No update in a while, but this would be a HUGELY useful feature :)

pingles:
We're not working on it (I'm afraid). However, I'd still love it to be added- we're in a reasonable place to support it given the API changes we made in v3.

If anyone else would like to pick it up I'd be happy to jump on a Zoom/Google Meet or something to give some crude guides to how it could be implemented.

PR #104, opened 2018-06-30
Validate Service Account Token of Agent

moofish32:
This is the first attempt at an incremental change focused on #91
 * Added the logic to review a token
 * Modified the proto messages
 * Small modifications to add configure objects

TODO: 
 * move tests to to test folder or should we move to more idiomatic go of <name>_test in folder?
 * Update rbac role in deploy to add system:auth-delegator for token review permissions
 * Update docs to describe new config params


moofish32:
ahh!! sorry for the vendor change, I'll be attempting to cleanly remove that in the next day

pingles:
Thanks @moofish32.

Sorry I've not done much with this yet- we started talking about changing some API stuff separately so I've started on that also. The changes are on the api-v3 branch. Once I've got those tested and integrated I'll try and go through your PR to merge.

moofish32:
@pingles -- can we get a PR started for your api-v3? I would like to add comments and we can likely close this because the v3 change is large enough that merging this will become difficult at best. First though a big +1 to the direction for moving the control of policy check back inside the server and reducing the client calls.

Joseph-Irving:
closing due to inactivity

Issue #108, opened 2018-07-04
Add some end-to-end tests


pingles:
It'd be great to have some automated tests that stand up Kiam (agent and server) against fakes for the AWS API and Kubernetes API servers to prove the behaviour of the whole system (and catch stuff any current unit tests miss).

Issue #105, opened 2018-07-03
Validate Agent and Pod are on the same Node


moofish32:
This is an improvement to reduce the ability to further exploit credentials when a node is compromised. I acknowledge node compromise is already a significant event. 

However GRPC has Peer information available in [context](https://godoc.org/google.golang.org/grpc/peer) so this should be a relatively light lift since we have the Pod information already.

This will enable teams to further put defense in depth by controlling pod placement, again with the k8s constraints this is not a declarative solve, but an improvement.

mattmb:
This would be great! I think validating that requesting IP of the agent is running the Pod it is asking for by checking the k8s API agrees would make things much more secure.

In my eyes, this is the major security benefit of the agent/server model. I guess checking that the Pod is running is a good start. But if a compromised agent means that I can get any credentials for anything that is running in the cluster then it's not a lot better than giving the agents the ability to do `sts:assume...`

geekofalltrades:
This seems very closely related to, or maybe even the same as, one of my main concerns with kiam. I want a kiam server to not hand out role credentials to a kiam agent if the node that that kiam agent is running on doesn't have a pod that's annotated for that role.

The more I turn this issue over in my head, the more I think it's the same as what I want, just worded slightly differently. I think I had in my head that the server would consider the role being asked for, and check the list of Pods on that agent's node to see if it should be allowed; but this issue is saying instead that the server would consider the Pod that is asking for the role through an agent, and verify that that agent and that Pod are on the same node.

I think those are functionally the same? Does that sound the same to everyone else?

Will a kiam server not hand out role credentials unless a pod name is specified in the request? Or is just the name of the role sufficient for credentials to be returned? I'll be reviewing the source code over the next few days and expect I'll discover this myself soon.

solarkennedy:
It looks like, as long as you have a cert and can connect to the kiam GRPC service, the only 2 pieces of information you need are an [IP and Role](https://github.com/uswitch/kiam/blob/92855a423e1cb9c3b54d275d297d63d95d809d43/proto/service.proto#L15).

I agree it would be cool if when it looked up the [pod](https://github.com/uswitch/kiam/blob/92855a423e1cb9c3b54d275d297d63d95d809d43/pkg/server/server.go#L162) for that IP, it inspected the node somehow and correlated it from the incoming Peer.Addr in the context.

geekofalltrades:
This is important to my organization. We have some changes internally that implement it. However:

[...]

Issue #38, opened 2018-02-20
Customizing suffix of the assumed role ARN for auditing purpose?


mumoshu:
Hi! Thanks for starting & maintaining this great project 👍 

I recently realized that running `get-caller-identity` with the kiam-provided AWS credentials produce:

```
$ aws sts get-caller-identity --region ap-northeast-1
```

```json
{
    "Account": "$myaccountid",
    "UserId": "AROAICVHQ4GZUSQIQRRHY:kiam-kiam",
    "Arn": "arn:aws:sts::$myaccountid:assumed-role/my-k8s-service-role/kiam-kiam"
}
```
`my-k8s-service-role` is from the pod annotation and `kiam-kiam` seems to be coming from kiam.

I couldn't find whether it is hard-coded into kiam or not.
Anyway, it would be great if I could configure the part to e.g. `kiam-$mycluster-$ns-$pod_name` so that it looks `kiam-mycluster-kube-system-mytestpod` which would add more traceability via CloudTrail logs. More concretely, it would be nice if I could trace from CloudTrail logs which pod in which cluster/namespace called which AWS API.

mumoshu:
Seems like these are where it is set:
https://github.com/uswitch/kiam/blob/e809a1aeeda7b5066329e269606ea43fcf6a575b/pkg/aws/sts/cache.go#L53
https://github.com/uswitch/kiam/blob/bbbc49733609414f333aa06864727c7b38fb160f/pkg/server/server.go#L159
https://github.com/uswitch/kiam/blob/bbbc49733609414f333aa06864727c7b38fb160f/cmd/server/main.go#L56

So, ideally a template to the session flag like `--session ${cluster_name}-{{.PodNamespace}}-{{.PodName}}` executed with the metadata from the credentials-requestor pod would be great.

pingles:
As you say- it's configurable with `--session` but currently kiam will request credentials (and cache) per-role, so the session would be tied to the originally requesting Pod (if you have the same role used by multiple pods it might be misleading) so given it's current implementation it'd mostly be useful for cluster name.

We re-use credentials as we often have many replicas of deployments/jobs etc. that use the same credentials and this helps reduce the number of AWS calls we make.

I guess another option would be to add an additional annotation that would name the session- then have credentials cached using the role name and session name (so pods in the same deployment/job etc. would share the same?).

pingles:
Also @mumoshu sorry for only just noticing this issue now :)

pingles:
[...]

PR #109, opened 2018-07-04
v3 API and refactorings

pingles:
This change was focused on changes to move to a new Agent/Server interaction ahead of v3. As part of doing this I also refactored/changed a few other interfaces to streamline stuff a little.

## API Changes
This adds a new Server API call: `GetPodCredentials` that allows credentials to be obtained by the server given an IP address. Previously the agent would make three separate calls:

1. Find the Pod (and it's role)
2. Check the Pod can assume the role
3. Get credentials for the role

This collapses calls together inside the server. This makes it easier to deliver #14, as well as potentially moving to other mechanisms for agents to request credentials (like the service account token introduced in #104.

## Testing
I've also broken apart the metadata server tests and moved all tests into the package directories, the Drone build also now runs with coverage stats to make it easier to figure out what parts of the system need tests added.

## End to End Testing
Although the tests are in better shape there's still things that aren't covered and would be nice to have some end-to-end tests to prove the whole- for now I've parked that and created #108 to track separately.

## Review and Request for Comments
I'd like some review/thoughts on the protos in particular. For now I've added new API calls and left the existing, so that the same server could support both new (3.x) and old (2.x) agents. However, given Kubernetes makes it easy to run servers side-by-side (only the agents have a hard requirement on running one per node) another option is to remove the API calls and have agents talk to different server services.


Makefile
```diff
@@ -19,12 +19,15 @@ proto/service.pb.go: proto/service.proto
 	go get -u -v github.com/golang/protobuf/protoc-gen-go
 	protoc -I proto/ proto/service.proto --go_out=plugins=grpc:proto
 
-test: $(SOURCES)
-	go test test/unit/*_test.go
-	go test test/functional/*_test.go
+test: $(shell find . -name '*.go')
```
sambooo: `$(SOURCES)`!

pkg/aws/metadata/handler_credentials.go
```diff
@@ -20,19 +20,19 @@ import (
...
 	"net/http"
 	"time"
 )
 
 type credentialsHandler struct {
-	roleFinder          k8s.RoleFinder
-	credentialsProvider sts.CredentialsProvider
-	clientIP            clientIPFunc
-	policy              server.AssumeRolePolicy
+	clientIP clientIPFunc
```
sambooo: The name `clientIP` is a bit misleading in a few places. We're not passing around an ip address but rather an ip resolution function. Not massively important though.

pkg/aws/metadata/handler_credentials.go
```diff
@@ -20,19 +20,19 @@ import (
...
+	client   server.Client
+}
+
+func (c *credentialsHandler) Install(router *mux.Router) {
```
sambooo: I like this approach to splitting up the handlers 👍

pkg/aws/metadata/handler_credentials.go
```diff
@@ -50,34 +50,10 @@ func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter,
...
-	if err != nil {
-		return http.StatusInternalServerError, fmt.Errorf("error checking assume role permissions: %s", err.Error())
-	}
-
-	if !decision.IsAllowed() {
-		return http.StatusForbidden, fmt.Errorf("assume role forbidden: %s", decision.Explanation())
-	}
-
-	credentials, err := c.credentialsForRole(ctx, requestedRole)
+	credentials, err := c.fetchCredentials(ctx, ip, requestedRole)
```
sambooo: There's no more code that can be deleted from this handler without breaking it. Good stuff.

pkg/testutil/server/client.go
```diff
@@ -0,0 +1,65 @@
+package server
+
+import (
+	"context"
+	"github.com/uswitch/kiam/pkg/aws/sts"
+)
+
+// StubClient returns fake server responses
+type StubClient struct {
```
sambooo: 👍 to this idea

pkg/aws/metadata/handler_credentials.go
```diff
@@ -90,15 +66,14 @@ func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter,
 	return http.StatusOK, nil
 }
 
-func (c *credentialsHandler) credentialsForRole(ctx context.Context, role string) (*sts.Credentials, error) {
+func (c *credentialsHandler) fetchCredentials(ctx context.Context, ip, requestedRole string) (*sts.Credentials, error) {
```
moofish32: this is a great improvement and reduces the ability of compromised client to just arbitrarily request roles.

pkg/aws/metadata/handler_role_name_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/uswitch/kiam/pkg/server"
+	st "github.com/uswitch/kiam/pkg/testutil/server"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+)
+
+func TestReturnRoleWhenClientResponds(t *testing.T) {
+	r, _ := http.NewRequest("GET", "/latest/meta-data/iam/security-credentials/", nil)
```
moofish32: All of these tests use `/latest/meta-data/iam/security-credentials/`. However, when you curl on an ec2 instance the `/latest/meta-data/iam/security-credentials` in at least some situations you don't get `200`:

```
curl -v 169.254.169.254/latest/meta-data/iam/security-credentials
* About to connect() to 169.254.169.254 port 80 (#0)
*   Trying 169.254.169.254...
* Connected to 169.254.169.254 (169.254.169.254) port 80 (#0)
> GET /latest/meta-data/iam/security-credentials HTTP/1.1
> User-Agent: curl/7.29.0
> Host: 169.254.169.254
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 301 Moved Permanently
< Location: http://169.254.169.254/latest/meta-data/iam/security-credentials/
< Content-Length: 0
< Connection: close
< Date: Mon, 09 Jul 2018 21:12:44 GMT
< Server: EC2ws
```
I don't think the SDK actually uses this endpoint, but should we make this match?

pkg/future/future_test.go
```diff
@@ -11,17 +11,16 @@
 // WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 // See the License for the specific language governing permissions and
 // limitations under the License.
-package kiam
+package future
```
moofish32: Is there something really unique about this future package? I know Capital One has this one: https://github.com/capitalone/go-future-context, I'm sure there are others. I just don't know if there is a specific reason the team wants to maintain this one. I agree it's a small amount of code right now.

pkg/server/policy.go
```diff
@@ -107,7 +107,7 @@ func (f *forbidden) Explanation() string {
 }
 
 func (p *RequestingAnnotatedRolePolicy) IsAllowedAssumeRole(ctx context.Context, role, podIP string) (Decision, error) {
-	pod, err := p.pods.GetPodByIP(ctx, podIP)
+	pod, err := p.pods.GetPodByIP(podIP)
```
moofish32: This is a point where I'm curious if we should validate the client IP (kiam client making the request) and the HostIp in the [PodStatus](https://godoc.org/k8s.io/api/core/v1#PodStatus) are a match? This would prevent being able to use a compromised host to provide IAM credentials for pods that could not be scheduled on the host. Probably a future enhancement but wanted to pose the question.

pkg/server/server.go
```diff
@@ -72,6 +75,41 @@ type KiamServer struct {
...
+		logger.Errorf("error checking policy: %s", err.Error())
+		return nil, err
+	}
+
+	if !decision.IsAllowed() {
+		logger.WithField("policy.explanation", decision.Explanation()).Errorf("pod denied by policy")
+		return nil, ErrPolicyForbidden
+	}
+
+	creds, err := k.credentialsProvider.CredentialsForRole(ctx, req.Role)
```
moofish32: So another potentially breaking change will come here when we want to specify the session name by pod ID. I don't know if that is important enough to roll in here or wait until later. I think the usefulness of a constant session name is fairly low.

pkg/server/server.go
```diff
@@ -181,7 +224,7 @@ func NewServer(config *Config) (*KiamServer, error) {
 		arnResolver,
 	)
 	server.credentialsProvider = credentialsCache
-	server.manager = prefetch.NewManager(credentialsCache, server.pods, server.pods)
+	server.manager = prefetch.NewManager(credentialsCache, server.pods)
```
moofish32: One more potential big change to consider is moving the credential cache into the pod cache. Logically if you want unique credentials per pod, then maintaining two caches is less useful. If this is potentially a major version change to implement and it's high enough priority, should we consider grouping the change here?

pkg/aws/metadata/handler_role_name_test.go
sambooo: I believe you're getting a 301 rather than 200 because you didn't including the trailing slash. See [this Google blog post](https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html): 
> don’t both return a 200 response code, but that one version redirects to the other.

pkg/server/policy.go
sambooo: It's only the agent that directly receives requests from clients, after which point the ip address forwarded by the agent is used to identify the pod. Are you suggesting the agent also pass its host ip to the server in this api call?

I suppose it does make sense to check that as a pod shouldn't hit another host's agent. @pingles would supporting this be another major bump or just an extension of the current request struct with graceful fallback?

pkg/aws/metadata/handler_role_name_test.go
pingles: Yeah I think @moofish32 was pointing out that the metadata server they tested on 301'ed, rather than returning credentials. Kiam currently supports either form and serves 200s on both.

@moofish32 I seem to remember others mentioning this stuff varied between instance generations (so it changed between m4 and m5 for example).

Do you remember the instance type you tested this on?

Either way I suspect we should probably add an issue to change it so that it behaves the same way- 301'ing rather than successfully working on both?

pkg/server/server.go
pingles: There's a separate issue open for being able to name sessions #38. 

The signature for `CredentialsForRole` will change but the only published API is the gRPC server's `GetPodCredentials` which should tolerate changes to reference pods by IP/ID etc.

So for now I'd leave this and pick it up as part of #38.

pkg/server/server.go
pingles: I'd defer picking it up until we do #38. I don't think the change would need a major version change to implement. I'd like us to sort #38 before we release v3 though- so we can still review the design for different pod and credential caches, the way sessions are named etc.

pkg/server/policy.go
pingles: @moofish32 I like the suggestion as an additional security enhancement (I assume this means check that the pod's host matches the agent's client IP?)

If so- could you open a new issue to track it please? I'm not sure if we'd commit to adding for v3 but I don't think it'd change the API as-is (the Pod IP is passed to `GetPodCredentials` which would allow for additional policy to check against the `PodStatus`).

moofish32: @sambooo -- we have both actually. The agent passes the IP of the pod and [GRPC](https://godoc.org/google.golang.org/grpc/peer) enables us to track the agent IP. The issue is tracked as #105

pkg/aws/metadata/handler_role_name_test.go
pingles: I've now added #116 to track this issue separately.

PR #110, opened 2018-07-04
Restrict metadata routes

### LINKED ISSUE #60: Block Access to Other Metadata
Kube2iam has a feature to block access to the rest of the metadata from the iptables level with a simple flag: https://github.com/jtblin/kube2iam#iptables

It seems like kiam did this by accident at one point https://github.com/uswitch/kiam/issues/3 and was undone. Anyway to make this optional. As we would like to block access to the metadata for our pods.

sambooo:
This set of changes adds a flag to `kiam agent` which allows the user to provide a regular expression that a request path must match to be proxied to the metadata service. The parameter doesn't affect any requests that are handled by kiam directly.

By default its value is set to `.*` which means all requests are allowed.

Resolves #60


pingles:
Thanks for this @sambooo. If you don't mind let's take a look at what we can do to update the testing to bring in line with the (unfortunately mammoth, sorry) changes I made in #109.

sambooo:
> update the testing to bring in line with ... #109.

Sure, let's gets that PR merged and I'll rebase

sambooo:
I've just pushed having rebased onto `api-v3`. Updated tests still to come.

cmd/kiam/agent.go
```diff
@@ -45,6 +47,7 @@ func (cmd *agentCommand) Bind(parser parser) {
 
 	parser.Flag("port", "HTTP port").Default("3100").IntVar(&cmd.port)
 	parser.Flag("allow-ip-query", "Allow client IP to be specified with ?ip. Development use only.").Default("false").BoolVar(&cmd.allowIPQuery)
+	parser.Flag("allowed-routes", "Proxy routes matching this regular expression").Default(".*").RegexpVar(&cmd.allowedRoutes)
```
ewbankkit: Should the flag be plural - `routes` - if we only allow it to be specified once?

pkg/aws/metadata/handler_proxy.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (p *proxyHandler) Handle(ctx context.Context, w http.ResponseWriter, r *http.Request) (int, error) {
+	route := mux.Vars(r)["path"]
+	if p.allowedRoutes.MatchString(route) {
+		writer := &teeWriter{w, http.StatusOK}
+		p.reverseProxy.ServeHTTP(writer, r)
+		return writer.status, nil
+	} else {
+		return http.StatusForbidden, fmt.Errorf("request blocked by allowedRoutes pattern %q: %s", p.allowedRoutes, route)
```
ewbankkit: Is `403` the correct HTTP status code to return?
From [RFC 7231](https://tools.ietf.org/html/rfc7231#section-6.5.3):
>  An origin server that wishes to "hide" the current existence of a
> forbidden target resource MAY instead respond with a status code of
> 404 (Not Found).

cmd/kiam/agent.go
sambooo: it's a regular expression which can match many routes

pkg/aws/metadata/handler_proxy.go
sambooo: the request is not allowed by (does not match) the `allowed-routes` regular expression, so the request is forbidden by the configuration. which status code should be returned in this case?

ewbankkit: I don't think there's a "correct" answer. Either would be valid.

cmd/kiam/agent.go
pingles: Perhaps then we could rename to specify that it's a regexp? So, change it to:

```
kiam server --whitelist-route-regexp=.*
```

Thoughts @sambooo and @ewbankkit? Agree about the plural name though- we don't support multiple regexes so it'd be good to name it singularly to avoid confusion.

pkg/aws/metadata/handler_proxy.go
pingles: Given that RFC maybe we should favour 404?

We should still log that it was blocked via a warning because of the path regexp (so that its easy for operators to record/debug) but 404 feels like it leaks less information to clients.

@uswitch/cloud any other thoughts? Feels pretty close to a coin toss though 😄

sambooo: I have no issues with changing the status code here since I went with my gut and `MAY` means it's up to us. Flip that coin.

cmd/kiam/agent.go
ewbankkit: Naming is the hardest problem 😄.
No particular opinion on this one but I think some documented examples would be good.

pingles:
Awesome. Last piece of the puzzle is that we want to require metadata API routes to be whitelisted, rather than permitting all by default (this was originally raised in #60).

I think once that's changed we're good to merge!

pkg/aws/metadata/handler_credentials.go
```diff
@@ -17,18 +17,19 @@ import (
...
 	"github.com/uswitch/kiam/pkg/server"
-	"net/http"
-	"time"
 )
 
 type credentialsHandler struct {
-	clientIP clientIPFunc
-	client   server.Client
+	client     server.Client
+	ipResolver clientIPFunc
```
pingles: I'd rename `ipResolver` to `getClientIP`

pkg/aws/metadata/handler_proxy_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+
+	"github.com/gorilla/mux"
+)
+
+func TestProxyDefaultBlacklisting(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+
+	allowedRoutes := regexp.MustCompile("^$")
```
pingles: If we're not going to move this statement inside `newProxyHandler` I'd like the expression stored in a const for something like `WhitelistDefaultBlockEverything` so that it can be referenced in both the package and test.

pkg/aws/metadata/handler_proxy_test.go
```diff
@@ -0,0 +1,109 @@
...
+	if rr.Code != http.StatusNotFound {
+		t.Error("unexpected status", rr.Code)
+	}
+}
+
+func TestProxyFiltering(t *testing.T) {
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*5)
+	defer cancel()
+
+	allowedRoutes := regexp.MustCompile("foo.*")
```
pingles: Passing in a non-default regexp is fine though

pkg/aws/metadata/handler_role_name.go
```diff
@@ -16,18 +16,19 @@ package metadata
...
 	"github.com/uswitch/kiam/pkg/server"
-	"net/http"
-	"time"
 )
 
 type roleHandler struct {
-	client   server.Client
-	clientIP clientIPFunc
+	client     server.Client
+	ipResolver clientIPFunc
```
pingles: As before, I'd prefer this to be named `getClientIP`.

pkg/aws/metadata/server.go
```diff
@@ -34,16 +36,18 @@ type Server struct {
...
-	MetadataEndpoint string
-	AllowIPQuery     bool
+	ListenPort           int
+	MetadataEndpoint     string
+	AllowIPQuery         bool
+	WhitelistRouteRegexp *regexp.Regexp
 }
 
-func NewConfig(port int) *ServerConfig {
+func NewConfig(port int, allowIPQuery bool, whitelistRouteRegexp *regexp.Regexp) *ServerConfig {
```
pingles: I'd rather we didn't expand NewConfig to include all options.

To be honest, I'd prefer it didn't specify a port but set good defaults for everything and then clients set individual fields.

Otherwise, as the struct grows we're just adding more and more params to the function which isn't nice.

pingles: Maybe we should rename the func to `DefaultConfig` too.

Issue #116, opened 2018-07-12
Metadata server: canonical redirect for credentials


pingles:
Initially reported by @moofish32.

Currently Kiam will serve successfully on both `/latest/meta-data/iam/security-credentials` and `/latest/meta-data/iam/security-credentials/`.

@moofish32 noticed that recent instances perform a 301 to redirect:

```
curl -v 169.254.169.254/latest/meta-data/iam/security-credentials
* About to connect() to 169.254.169.254 port 80 (#0)
*   Trying 169.254.169.254...
* Connected to 169.254.169.254 (169.254.169.254) port 80 (#0)
> GET /latest/meta-data/iam/security-credentials HTTP/1.1
> User-Agent: curl/7.29.0
> Host: 169.254.169.254
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 301 Moved Permanently
< Location: http://169.254.169.254/latest/meta-data/iam/security-credentials/
< Content-Length: 0
< Connection: close
< Date: Mon, 09 Jul 2018 21:12:44 GMT
< Server: EC2ws
```

It'd be good to change the agent metadata proxies to behave in the same way.

pingles:
This is now merged. Thanks for reporting @moofish32.

It'll be in the v3 release (aiming for a release candidate in the next week or so).

dblackdblack:
Ugh.  unfortunately this is not how kube2iam works (it returns a 200 code for both the trailing backslash and non trailing backslash URLs).  Another minor roadbump for switching away from kube2iam.

PR #121, opened 2018-07-16
metadata server: redirect security-credentials when no trailing slash

pingles:
This addresses #116: canonical redirect.

Previously we served on both:

* `/{version}/meta-data/iam/security-credentials/`
* `/{version}/meta-data/iam/security-credentials`

This changes so that the latter is redirected to the former.


pingles:
@uswitch/cloud could someone review please

PR #122, opened 2018-07-16
Record pod events for all server calls

### LINKED ISSUE #14: Record IAM errors as events

sambooo:
Resolves #14

### DIFF
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -97,7 +97,7 @@ func (k *KiamServer) GetPodCredentials(ctx context.Context, req *pb.GetPodCreden
 	if !decision.IsAllowed() {
 		logger.WithField("policy.explanation", decision.Explanation()).Errorf("pod denied by policy")
 		k.recordEvent(pod, v1.EventTypeWarning, "KiamRoleForbidden",
-			fmt.Sprintf("failed assuming role: %s", req.Role))
+			fmt.Sprintf("failed assuming role %q: %s", req.Role, decision.Explanation()))
 		return nil, ErrPolicyForbidden
 	}
 
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -32,11 +32,11 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/tools/record"
-	"k8s.io/client-go/tools/reference"
 )
 
 // Config controls the setup of the gRPC server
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -96,15 +96,20 @@ func (k *KiamServer) GetPodCredentials(ctx context.Context, req *pb.GetPodCreden
 
 	if !decision.IsAllowed() {
 		logger.WithField("policy.explanation", decision.Explanation()).Errorf("pod denied by policy")
+		k.recordEvent(pod, v1.EventTypeWarning, "KiamRoleForbidden",
```
pingles: In this event we should also be able to use `decision.Explanation()` to include more information about why it was rejected.

pkg/server/server.go
```diff
@@ -96,15 +96,20 @@ func (k *KiamServer) GetPodCredentials(ctx context.Context, req *pb.GetPodCreden
...
 
 	creds, err := k.credentialsProvider.CredentialsForRole(ctx, req.Role)
 	if err != nil {
 		logger.Errorf("error retrieving credentials: %s", err.Error())
+		k.recordEvent(pod, v1.EventTypeWarning, "KiamCredentialError",
+			fmt.Sprintf("failed retrieving credentials: %s", err))
 		return nil, err
 	}
 
+	k.recordEvent(pod, v1.EventTypeNormal, "KiamCredentialIssued", "issued credentials")
```
pingles: I'd want to check our metrics but I have a feeling this could get called a _lot_. Does the Event API already guard against logging the same message over and over? Don't want to do something that DDoS the API server/etcd :)

pkg/server/server.go
```diff
@@ -139,15 +144,10 @@ func (k *KiamServer) GetPodRole(ctx context.Context, req *pb.GetPodRoleRequest)
...
-	ref, err := reference.GetReference(scheme.Scheme, pod)
-	if err != nil {
-		logger.Errorf("error getting reference for pod %q: %s", pod.Name, err)
-		return nil, err
-	}
 
 	logger.WithField("pod.iam.role", role).Infof("found role")
-	k.eventRecorder.Event(ref, v1.EventTypeNormal, "KiamRoleFound",
-		fmt.Sprintf("Role: %q found for pod: %q", role, pod.Name))
+	k.recordEvent(pod, v1.EventTypeNormal, "KiamRoleFound",
```
pingles: As above, I'd want to know we could do this without risk of DDoS'ing the API server and etcd- I think these endpoints will get called a lot.

pkg/server/server.go
sambooo: I suppose neither of these `Normal` events is necessary now I think about it. The fact the pod is running without issue shows things are normal.

pingles: I guess if it's possible to _only_ have credentials, or an error, then its safe to remove.

One upside to include is that if people don't see any errors, or successful messages they'll know that the annotation was misspelt/missing etc.

pingles:
Nice, thanks @sambooo.

Could you update the RBAC configuration in `/deploy/server-rbac.yaml` please? It'll need additional permissions to write events.

pingles:
Then I think we're in a place to merge and close #14, happy days!

PR #124, opened 2018-07-18
Instrument gRPC server and client

### LINKED ISSUE #114: Improve gRPC telemetry integration
#109 removed a bunch of the Telemetry integration (removing `TelemetryClient` and `TelemetryServer`).

This is to ensure that similar/same metrics are captured (something to time things within the server and the same operations as observed by the client within the agent process).

This also affects #113.

sambooo:
This change adds [grpc-prometheus](https://github.com/grpc-ecosystem/go-grpc-prometheus) metrics to both the Kiam server and agent. I also removed an unused dependency on the way.

Resolves #114


sambooo:
Coverage is down in this commit as none of our tests use the gRPC client or server directly at the moment. That can be ignored for now.

pingles:
It'd be great to get the gRPC stuff covered through #108. Happy to postpone for now and hopefully pick back up through that.

Gopkg.toml
```diff
@@ -37,10 +37,6 @@
   name = "github.com/coreos/go-iptables"
   version = "0.3.0"
 
-[[constraint]]
-  name = "github.com/fortytw2/leaktest"
```
pingles: Is this removed because we no longer use it?

Maybe I'll create a separate issue to make sure we add leaktest back to all tests- it picked up a few things in the early days.

PR #131, opened 2018-08-02
Idiamond metrics

pingles:
This builds upon the work by @idiamond-stripe to improve handling of Prometheus and StatsD metrics.

It renames some of the metrics @idiamond-stripe introduced to bring them closer in line with the naming recommendations (`_total` only for counters etc.).

Currently gRPC Client metrics (those exported within the agent) don't report correctly- once that's fixed this can be merged.


pingles:
If someone from @uswitch/cloud could take a look and see what you think- I think this should be the last thing needed before we push out an RC and start testing in some of our clusters.

docs/metrics.md
```diff
@@ -0,0 +1,72 @@
+# Metrics
+
+Kiam exports both Prometheus and StatsD metrics to determine the health of the
+system, check the timing of each RPC call, and monitor the size of the
+credentials cache. By default, Prometheus metrics are exported on
+`localhost:9620` and StatsD metrics are sent to `127.0.0.1:8125`. StatsD
+metrics are not aggregated and flushed every 100ms.
```
Joseph-Irving: `are not aggregated and flushed every 100ms.` I think that should be "are not aggregated and **are** flushed every 100ms"

Issue #133, opened 2018-08-03
Readinessprobe panics with latest kiam source


roffe:
Tried building kiam from source as i wanted the new metrics stuff ( working on a dashboard )

But now kiam server is failing readiness probes

```
 Warning  Unhealthy  12s                kubelet, ip-10-0-118-76.eu-west-1.compute.internal  Readiness probe failed: panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x85fbf2]

goroutine 1 [running]:
github.com/uswitch/kiam/vendor/gopkg.in/alexcesaro/statsd%2ev2.(*Client).skip(0x0, 0x2ac8df26)
  /Users/jbk/go/src/github.com/uswitch/kiam/vendor/gopkg.in/alexcesaro/statsd.v2/statsd.go:87 +0x22
github.com/uswitch/kiam/vendor/gopkg.in/alexcesaro/statsd%2ev2.(*Client).Timing(0x0, 0x13b665e, 0x12, 0x11b1100, 0xc420499bf8)
  /Users/jbk/go/src/github.com/uswitch/kiam/vendor/gopkg.in/alexcesaro/statsd.v2/statsd.go:105 +0x2f
github.com/uswitch/kiam/vendor/gopkg.in/alexcesaro/statsd%2ev2.Timing.Send(0xbed133c3eac04d7a, 0x3b6ebd6, 0x1d01280, 0x0, 0x13b665e, 0x12)
  /Users/jbk/go/src/github.com/uswitch/kiam/vendor/gopkg.in/alexcesaro/statsd.v2/statsd.go:132 +0xaa
github.com/uswitch/kiam/pkg/server.(*KiamGateway).Health(0xc42000a100, 0x150e520, 0xc420134420, 0xc420406060, 0x2, 0x0, 0x0)
  /Users/jbk/go/src/github.com/uswitch/kiam/pkg/server/gateway.go:156 +0x150
main.(*healthCommand).Run.func1(0xc42000a140, 0xc42000a140)
  /Users/jbk/go/src/github.com/uswitch/kiam/cmd/kiam/health.go:55 +0x4e
github.com/uswitch/kiam/vendor/github.com/cenkalti/backoff.RetryNotify(0xc42000a120, 0x14f7420, 0xc42000a140, 0x0, 0xc42000a100, 0x14f7420)
[...]

PR #134, opened 2018-08-03
don't push statsd metrics if it's disabled

roffe:
adressing https://github.com/uswitch/kiam/issues/133

### DIFF
--- a/cmd/kiam/agent.go
+++ b/cmd/kiam/agent.go
@@ -79,12 +79,19 @@ func (opts *agentCommand) Run() {
 	ctxGateway, cancelCtxGateway := context.WithTimeout(context.Background(), opts.timeoutKiamGateway)
 	defer cancelCtxGateway()
 
-	gateway, err := kiamserver.NewGateway(ctxGateway, opts.serverAddress, opts.serverAddressRefresh, opts.caPath, opts.certificatePath, opts.keyPath)
+	var statsd bool
+	if opts.telemetryOptions.statsD != "" {
+		statsd = true
+	} else {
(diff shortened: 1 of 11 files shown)


roffe:
I've done some testing today with the new code to ensure stuff is working as it should.

I can delete the server with running agents and the agents will pick up on the new server ip that emerges and direct the requests there within seconds

Noteworthy might be that i'm running CoreDNS with the default 5 sec TTL on cluster records,
and I have ` - --gateway-timeout-creation=1s` on agent & livliness & readiness check for server as 50msec can be to low with the new logic i put in place in server.NewGateway.

I use grpc.DialContext instead of Dial since as far as i understand WithBlock does not work with Dial

added opts to make sure conn is up and we have received a setting frame before handing it back
grpc.WithBlock()
grpc.WithWaitForHandshake()

WithBlock returns a DialOption which makes caller of Dial blocks until the underlying connection is up. Without this, Dial returns immediately and connecting the server happens in background.

WithWaitForHandshake blocks until the initial settings frame is received from the server before assigning RPCs to the connection.

pingles:
Thanks. I'd prefer to avoid the proliferation of the StatsD stuff- @idiamond-stripe worked on replacing much of the metric stuff recently. Is there another way to do this without having to pass in parameters everywhere?

@roffe I'll try and look through this in more detail at work tomorrow- I'd definitely like to take in the gRPC related stuff asap from a cursory look through, thanks for all that.

roffe:
new img roffe/kiam-git:v2.8-37-gf4b89b2

Issue #135, opened 2018-08-08
Assumed role arn not trusted by instance role


kevtaylor:
Kiam version 2.7

When a role is created as an 'override' to the generated version in kube-aws, there are assume errors as:

```
{"level":"error","msg":"error requesting credentials: AccessDenied: User: arn:aws:sts::034324643013:assumed-role/us-west-2-md-Controller/i-084d38eaacc983616 is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::034324643013:role/us-west-2-md-Kiam\n\tstatus code: 403, request id: 01b4ba14-9af0-11e8-b9ba-c1a73e9c177a","pod.iam.role":"us-west-2-md-external-dns","time":"2018-08-08T09:47:04Z"}
```
In this example the `us-west-2-md-Controller` is the one that is created by kube-aws when the cluster spins (we specify this role name in the cluster.yaml configs)

And we want to replace the kiam server with `- --assume-role-arn=arn:aws:iam::034324643013:role/us-west-2-md-Kiam`

pingles:
Ah... I've just reread the error- it looks like your instance `i-084d38eaacc983616` is attempting to assume `arn:aws:iam::034324643013:role/us-west-2-md-Kiam` and isn't being permitted- does your `us-west-2-md-Kiam` role have sufficient assume role policy to permit it? 

I initially thought it was Kiam not attempting to assume the role you'd specified in `--assume-role-arn` but it looks like it might be a trust policy issue?

kevtaylor:
So, if I add an AssumeRole to the controller iam ie:
```
{
      "Sid": "",
      "Effect": "Allow",
      "Principal": {
        "AWS": "arn:aws:iam::034324643013:role/us-west-2-md-Controller"
      },
      "Action": "sts:AssumeRole"
    }
```

That works, but that sort of defeats the object because I want to create the roles independently of the cluster. Is there a more 'generic' way?

tombooth:
Can you expand on what you mean by 'creating roles independently of the cluster' and a 'more generic way'?

At the moment the Kiam has to use AWS roles as a way of allowing operators to restrict only certain roles to be available to the Kubernetes cluster. It assumes this 'cluster identity' role on start and as specified by `--assume-role-arn`. This is very corse and relies on Kiam restricting the which pods can ask for which roles, through use of namespace restrictions. 

[...]

PR #139, opened 2018-08-09
Updated IAM documentation

pingles:
I've started working on improving the IAM documentation to be clearer about how to configure roles and policies.

It's come up a few times recently (#135 most recently) and is something we've been wanting to improve for a long while #8.

It's a work-in-progress so this PR is just so that other people can see and help improve it too!

### DIFF
--- a/docs/IAM.md
+++ b/docs/IAM.md
@@ -45,10 +45,10 @@ The example below is expressed using
 to Terraform configuration and should help explain how AWS IAM resources are
 connected. 
 
-```ruby
-aws_iam_role :server_node, {
-  name: "server_node",
-  assume_role_policy: <<-EOF
+```
+resource "aws_iam_role" "server_node" {
+  name = "server_node"
+  assume_role_policy = <<EOF
 {
   "Version": "2012-10-17",
   "Statement": [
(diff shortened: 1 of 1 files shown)


pingles:
@Joseph-Irving fixed :-)

docs/IAM.md
```diff
@@ -1,13 +1,40 @@
-# IAM Policy
+# IA
```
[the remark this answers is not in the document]
pingles: yep... _sigh_

docs/IAM.md
```diff
@@ -1,13 +1,95 @@
...
+
+#### Server Node Policy
+This is the example policy that will allow the EC2 instance that runs the Server
+process can assume the server role. 
+
+The example below is expressed using
+[Terrafying](https://github.com/uswitch/terrafying) but is pretty close
+to Terraform configuration and should help explain how AWS IAM resources are
+connected. 
+
```
Joseph-Irving: why use terrafying instead of terraform? more people will be familiar with terraform

docs/IAM.md
```diff
@@ -1,13 +1,95 @@
...
+  policy: <<-EOF
+  {
+  "Version": "2012-10-17",
+  "Statement": [
+    {
+      "Effect": "Allow",
+      "Action": [
+        "sts:AssumeRole"
+      ],
+      "Resource": "arn:aws:iam:123456789012:role/kiam-server"
```
DewaldV: This needs to be the roles the server can assume. Maybe start with an example that has `["*"]`?
Alternatively `["arn:aws:iam:123456789012:role/my-pod-role"]`?

docs/IAM.md
pingles: Because that's what we use. Using it was just to help show how the roles, policies etc. attach- both languages (Ruby, HCL) should be readable.

docs/IAM.md
pingles: Well the node role is just to allow the Server process to obtain credentials to the `kiam-server` role which is the role then used to assume pod roles.

DewaldV: Ahh the chains of sts! I got it now. Nevermind me then. :)

PR #140, opened 2018-08-10
Example Grafana dashboard

roffe:
Example dashboard for Grafana with prometheus as datasource

### DIFF
--- a/docs/dashboard-prom.json
+++ b/docs/dashboard-prom.json
@@ -1,0 +1,2806 @@
+{
+    "__inputs": [],
+    "__requires": [
+      {
+        "type": "grafana",
+        "id": "grafana",
+        "name": "Grafana",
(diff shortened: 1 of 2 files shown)


docs/dashboard-prom.json
```diff
@@ -0,0 +1,2806 @@
...
+        "pointradius": 5,
+        "points": false,
+        "renderer": "flot",
+        "seriesOverrides": [],
+        "spaceLength": 10,
+        "stack": false,
+        "steppedLine": false,
+        "targets": [
+          {
+            "expr": "sum by (pod, handler) (delta(kiam_metadata_success_total[$interval]))",
```
rehevkor5, line 104: FYI these uses of `delta()` should have been `increase()` instead, as `delta()` does not understand counter resets and therefore can produce nonsensical negative values.

Issue #100, opened 2018-06-27
Reduce Docker build context size


sambooo:
We can avoid uploading the binary directly to the build host by instead building in a container and copying the result.

roffe:
Usage examples:

```
jbk/def (⎈ |staging.sgn.io:kube-system) git:(multistage) uswitch/kiam ➜  make
docker image build -t "quay.io/uswitch/kiam:multistage" .
Sending build context to Docker daemon  40.97MB
Step 1/9 : FROM golang:1.10-alpine3.8 as build
1.10-alpine3.8: Pulling from library/golang
8e3ba11ec2a2: Pull complete
8e6b2bc60854: Pull complete
3d20cafe6dc8: Pull complete
7a50139852bf: Pull complete
eda88e56e76e: Pull complete
Digest: sha256:875d980b77cea706bb05c954b485affeaf7fea054a5d65a4365141a38cea15c3
Status: Downloaded newer image for golang:1.10-alpine3.8
 ---> 34d3217973fd
Step 2/9 : RUN apk add --no-cache make
 ---> Running in 2306039be6da
fetch http://dl-cdn.alpinelinux.org/alpine/v3.8/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/v3.8/community/x86_64/APKINDEX.tar.gz
(1/1) Installing make (4.2.1-r2)
Executing busybox-1.28.4-r0.trigger
OK: 5 MiB in 15 packages
Removing intermediate container 2306039be6da
 ---> bb071837988d
Step 3/9 : WORKDIR /go/src/github.com/uswitch/kiam
 ---> Running in e4bb241fdbc6
Removing intermediate container e4bb241fdbc6
 ---> 95bf0da0e14f
Step 4/9 : ADD . .
 ---> 1b86e19269c9
Step 5/9 : RUN make proto/service.pb.go build-linux
 ---> Running in 2640c0e6a34e
make: 'proto/service.pb.go' is up to date.
GOARCH=amd64 GOOS=linux CGO_ENABLED=0 go build -o bin/kiam-linux-amd64 cmd/kiam/*.go
Removing intermediate container 2640c0e6a34e
 ---> 71d24e13ca5a
Step 6/9 : FROM alpine:3.8
3.8: Pulling from library/alpine
8e3ba11ec2a2: Already exists
Digest: sha256:7043076348bf5040220df6ad703798fd8593a0918d06d3ce30c6c93be117e430
[...]

PR #142, opened 2018-08-13
use multistage build Dockerfile

roffe:
Closes https://github.com/uswitch/kiam/issues/100

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+bin
+docs
+ci
+deploy
--- a/.drone.yml
+++ b/.drone.yml
@@ -16,10 +16,10 @@ pipeline:
     commands:
       - make bench
 
-  build:
-    image: golang:1.10
-    commands:
-      - make build-linux
+#  build:
+#    image: golang:1.10
+#    commands:
+#      - make build-linux
 
   docker:
     when:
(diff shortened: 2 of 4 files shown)


.drone.yml
```diff
@@ -16,10 +16,10 @@ pipeline:
     commands:
       - make bench
 
-  build:
-    image: golang:1.10
-    commands:
-      - make build-linux
+#  build:
```
pingles: Could you delete the commented section please

.dockerignore
```diff
@@ -0,0 +1,4 @@
+bin
```
pingles: Could this include `.git` also?

Thanks for adding this- be useful to exclude as much as we can.

Dockerfile
```diff
@@ -1,7 +1,10 @@
-FROM alpine:3.8
+FROM golang:1.10-alpine3.8 as build
+RUN apk add --no-cache make
+WORKDIR /go/src/github.com/uswitch/kiam
+ADD . .
+RUN make proto/service.pb.go build-linux
```
pingles: Could we simplify the Makefile so that the default is to build the binary and have it manage the rest please?

roffe:
something like this?

PR #152, opened 2018-09-06
Add support for static role aliases

alindeman:
Introduces a new command line option, `--role-alias`, which specifies a
mapping from a role alias to a full role name. A pod can then specify a
role alias via the `iam.amazonaws.com/role-alias` and have it resolved
to the full role.

This is primarily useful in cases where the full role differs between
environments, such as when IAM roles are created by CloudFormation.

A future improvement will be to dynamically look up and cache role
aliases in a config map or similar object.

Example:

```
kiam server --role-alias=external-dns=stack-12345-external-dns-98765
```


alindeman:
Whoops, didn't mean to open this quite yet. I'm still iterating on the idea.

Issue #153, opened 2018-09-07
The deploy files use the wrong path for the binary 


someword:
Using [these](https://github.com/uswitch/kiam/tree/master/deploy) the command is called `kiam` https://github.com/uswitch/kiam/blob/master/deploy/agent.yaml#L40 however in the recommended docker image the kiam binary does not exist but the agent,server and health binaries do.

> ❯ docker run -it --entrypoint=ls quay.io/uswitch/kiam:v2.8 /kiam
ls: /kiam: No such file or directory

>~/src/environment-deployment/kubernetes/templates/kiam CCSJP-3296*
❯ docker run -it --entrypoint=ls quay.io/uswitch/kiam:v2.8 /agent
/agent

>~/src/environment-deployment/kubernetes/templates/kiam CCSJP-3296*
❯ docker run -it --entrypoint=ls quay.io/uswitch/kiam:v2.8 /server
/server

>~/src/environment-deployment/kubernetes/templates/kiam CCSJP-3296*
❯ docker run -it --entrypoint=ls quay.io/uswitch/kiam:v2.8 /health
/health

pingles:
This changed between v2 and v3. v3 uses a single binary for all commands (which master refers to), and v2 uses the other commands.

Apologies for the confusion- if you want to get the deployment manifests for v2.7 you'll have to navigate via the releases- the tree has correct files.

PR #154, opened 2018-09-07
Set the correct path the binary in the image.  

someword:
This fixes https://github.com/uswitch/kiam/issues/153.


pingles:
I added a comment in #153 but this is because the documentation in master refers to v3 which changed kiam to produce a single binary with each command rather than separate binaries. Apologies for the confusion.

PR #157, opened 2018-09-21
Convert deploy to `kustomize` base

amitsaha:
Hi! I am looking to use `kiam` for my cluster and one of the first things I have to change  is the agent's CNI interface name. I think if we convert the deployment configuration that we have in the repo here to a [kustomize](https://github.com/kubernetes-sigs/kustomize) base configuration, operators of `kiam` can then use `overlays` to overrride any configuration. For example, see [here](https://github.com/amitsaha/kiam/tree/kustomize_overlays/deploy).

What do you think?


Joseph-Irving:
Thanks but I don't think we have any plans to support kustomize atm, people are welcome to use it if they so desire but I think offering helm and normal kubernetes manifests is enough for most

PR #158, opened 2018-09-25
Kops interface name in readme

word:
Adds kops interface name to readme


roffe:
@word Kops interface depends on CNI choice, maybe you should add a note like "When using default CNI" and link to https://github.com/kubernetes/kops/blob/master/docs/networking.md

Issue #36, opened 2018-02-14
failed to load system roots and no roots provided - TLS error


cdenneen:
```
{"generation.metadata":0,"level":"error","msg":"error warming credentials: RequestError: send request failed\ncaused by: Post https://sts.amazonaws.com/: x509: failed to load system roots and no roots provided","pod.iam.role":"arn:aws:iam::###########:role/chrisiamtest1","pod.name":"aws-cli3","pod.namespace":"default","pod.status.ip":"100.112.74.130","pod.status.phase":"Running","resource.version":"4849725","time":"2018-02-14T17:59:54Z"}

{"generation.metadata":0,"level":"error","msg":"error warming credentials: RequestError: send request failed\ncaused by: Post https://sts.amazonaws.com/: x509: failed to load system roots and no roots provided","pod.iam.role":"chrisiamtest1","pod.name":"aws-cli","pod.namespace":"default","pod.status.ip":"100.112.74.129","pod.status.phase":"Running","resource.version":"4849455","time":"2018-02-14T17:59:54Z"}

{"level":"error","msg":"error requesting credentials: RequestError: send request failed\ncaused by: Post https://sts.amazonaws.com/: x509: failed to load system roots and no roots provided","pod.iam.role":"arn:aws:iam::############:role/chrisiamtest1","time":"2018-02-14T18:00:54Z"}
```

I've tried with just using the role name and the full ARN in the pod deployment.
Can someone help me understand what this error means?
Is there documentation on how to specify the base-arn or is autodetect the best solution?

cdenneen:
I've logged into agent/server and the TLS secrets are in `/etc/kiam/tls` is they are supposed to be: https://github.com/uswitch/kiam/blob/master/docs/TLS.md

pingles:
@cdenneen The published docker image doesn't have any trusted CAs included so it's failing to interact with the AWS APIs. The container expects these to be in `/etc/ssl/certs`.

We use CoreOS Container Linux and so mount the CAs from the host into the container (so updates to the host propagate to the containers).

The relevant bits in the deployment manifests are:

* https://github.com/uswitch/kiam/blob/master/deploy/server.yaml#L21
* https://github.com/uswitch/kiam/blob/master/deploy/server.yaml#L44

Hope that helps. I'm going to close this for now but if something else happens please comment and we'll try to help.

cdenneen:
[...]

PR #159, opened 2018-09-27
Add ca-certificates to Docker image

knqyf263:
Related: https://github.com/uswitch/kiam/issues/36

`kiam-server` got the following error.
```
kiam-server-522xx kiam-server {"generation.metadata":0,"level":"error","msg":"error warming credentials: RequestError: send request failed\ncaused by: Post https://sts.amazonaws.com/: x509: failed to load system roots and no roots provided","pod.iam.role":"masters.k8s.dev.rafflesia.io","pod.name":"test-pod-b7bdb7747-dww85","pod.namespace":"default","pod.status.ip":"100.96.3.7","pod.status.phase":"Running","resource.version":"9615569","time":"2018-09-27T14:39:49Z"}
```

I know that this issue can be solved by mounting host's certificate.
https://github.com/uswitch/kiam/blob/ceb88e190bb9552f76a491dad41fcb5d0ba855b9/deploy/server.yaml#L21-L25

However, the certificate's path depends on host OS (also written in the comment)
It seems better to install the certificate in the docker image.
This eliminates the need to care about the host OS and configure the volume.

Before installing ca-certificates:
```
$ docker images kiam
REPOSITORY          TAG                 IMAGE ID            CREATED              SIZE
kiam                latest              b9290ba7f5f4        About a minute ago   50.1MB
```

After installing ca-certificates:
```
$ docker images kiam
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
kiam                latest              3b778a6f07b7        7 seconds ago       50.7MB
```

Image sizes are almost the same.

Thanks.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,6 @@ ADD . .
 RUN make
 
 FROM alpine:3.8
-RUN apk --no-cache add iptables
+RUN apk --no-cache add iptables ca-certificates
 COPY --from=build /go/src/github.com/uswitch/kiam/bin/kiam-linux-amd64 /kiam
 CMD []


pingles:
Thanks but I'm generally in favour of mounting and using the certificates from the host. I appreciate this makes deployments vary depending on host platform but it makes auto-updates easier. If you strongly feel different please open an issue and we can let more people discuss.

knqyf263:
@pingles Thank you for the reply!
I understand your claim.

However, even if the certificate is installed in the container, it is possible to mount the certificates from the host.
People who want to use the certificate of the host mount and override the certs directory.
People who doesn't want will not mount.

How about providing choices?

Issue #162, opened 2018-10-08
Change localhost references to 127.0.0.1


pingles:
I think it was @roffe that referenced https://pracucci.com/kubernetes-dns-resolution-ndots-options-and-why-it-may-affect-application-performances.html in another issue.

I'd suggest we change all the deployment manifest examples to reference `127.0.0.1` rather than `localhost` to avoid DNS during health checks.

cheeseandcereal:
The PR for this change https://github.com/uswitch/kiam/pull/163 caused me a LOT of headache trying to get cert-manager working with kiam. cert-manager (and other TLS certificate tools) don't all handle SAN extension for iPAddress correctly in order to get a working TLS cert for an IP address (namely 127.0.0.1). See [HERE](https://cabforum.org/guidance-ip-addresses-certificates/) for more info, but generally it's probably not best practice to be using TLS certs for IP addresses, even with a private CA.

Are we even sure that DNS lookups for localhost are even going to the kubernetes dns for lookup? In basically every linux configuration I've ever seen, the /etc/hosts file ALWAYS has precedence over a remote DNS lookup/search, and I've seen no evidence to suggest that `localhost` dns lookups ever leave the local pod in a k8s cluster, which would mean there's no sort of performance penalty for using `localhost` instead of `127.0.0.1`. Reverting to `localhost` would allow for regular `dNSName` fields to be used with the cert for the kiam server, which is vastly more supported.

It's especially misleading when the docs point out that 'when deploying to production please ensure you automate certificates using a tool like Vault, cert-manager etc.', when cert-manager literally doesn't work with ip address certificates, and it's not obvious you would ever need to change the local server-address for the liveliness/readiness probes, especially with no good errors even showing up in logs until you enable higher verbosity with GRPC_GO_LOG_VERBOSITY_LEVEL, and even then, it's only 'bad TLS cert' without any other explanation.

For these reasons, I personally think this change should be reverted unless it can be shown that dns lookups for localhost ACTUALLY make network calls (which I'm fairly certain they do not). And if so, it should at least be documented in https://github.com/uswitch/kiam/blob/master/docs/TLS.md that a cert must be generated with proper iPAddress SAN extension support if any ip's are going to be used for any `server-address` parameters in the deployment manifests.

hakamadare:
[...]

PR #163, opened 2018-10-10
fix(kiam-server-manifests): change localhost ref

tasdikrahman:
closes https://github.com/uswitch/kiam/issues/162


pingles:
Thanks @tasdikrahman and @roffe for the original reference.

PR #167, opened 2018-10-25
Missing v3 :-(

pingles:
I was messaged on Slack to point out that the whitelisting path changes added in #110 wasn't available in `master` and also missing from the `v3-rc1` image.

Having checked through history it looks like some later v3 changes were merged but into the previously merged `api-v3` branch but weren't then integrated to master. The missing commit appears to be https://github.com/uswitch/kiam/commit/afdb221cf63de4cdcf2e70f75660c25456103d75.

I cherry picked that commit and went through the merge conflicts; the result is these set of changes.

Note that, because this adds a change- it blocks metadata access by default- the default behaviour will be different. It also means the current `v3-rc1` is missing a key part of the v3 changes which will alter behaviour.

### DIFF
--- a/pkg/aws/metadata/handler_credentials.go
+++ b/pkg/aws/metadata/handler_credentials.go
@@ -24,12 +24,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/uswitch/kiam/pkg/aws/sts"
 	"github.com/uswitch/kiam/pkg/server"
-	"github.com/uswitch/kiam/pkg/statsd"
 )
 
 type credentialsHandler struct {
-	clientIP clientIPFunc
-	client   server.Client
+	client      server.Client
+	getClientIP clientIPFunc
 }
 
 func (c *credentialsHandler) Install(router *mux.Router) {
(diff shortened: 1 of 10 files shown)


pingles:
Given the confusion around this I'd especially appreciate some careful review from those that can @uswitch/cloud @roffe. 

It'll also mean that when we publish as `v3-rc2` there'll be breaking changes from `v3-rc1`.

pkg/aws/metadata/handler_credentials.go
```diff
@@ -39,16 +38,13 @@ func (c *credentialsHandler) Install(router *mux.Router) {
 func (c *credentialsHandler) Handle(ctx context.Context, w http.ResponseWriter, req *http.Request) (int, error) {
 	timer := prometheus.NewTimer(handlerTimer.WithLabelValues("credentials"))
 	defer timer.ObserveDuration()
-	if statsd.Enabled {
-		defer statsd.Client.NewTiming().Send("handler.credentials")
-	}
```
roffe: Intentional to drop the "Observed latency of credentials handler" metrics?

pingles: Nope, that wasn't intentional. :(

pingles:
For some reason I couldn't pull the latest repo changes and so the result of this merge was missing a bunch of changes. I'm going to re-do and submit as a new PR.

PR #183, opened 2018-11-23
Fix KIAM cache miss metric name on dashboard

max-lobur:
This metric is called `kiam_sts_cache_miss_total` as I see in Prometheus. I think you even have this bug on a dashboard screenshot in a doc (cache miss has no data there because metric name in json is incorrect). Here's my dashboard after I updated it:
image

### DIFF
--- a/docs/METRICS.md
+++ b/docs/METRICS.md
@@ -47,8 +47,8 @@ daemonset status from kube-state-metrics & container metrics from cAdvisor if av
 
 #### STS Subsystem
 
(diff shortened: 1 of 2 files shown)


max-lobur:
BTW really appreciate you for posting a dashboard. I wish all products have this!
https://twitter.com/max_lobur/status/1066026381128146947

max-lobur:
Metrics in prometheus as of `quay.io/uswitch/kiam:v3.0-rc1`

max-lobur:
If you want I can do the new screenshot for docs too.

PR #185, opened 2018-11-23
Grafana dashboard dehardcode namespace

max-lobur:
I don't deploy kiam to `kube-system`, instead I place it in own namespace. Current dashboard has `kube-system` hardcoded. 

This change adds a dropdown with the 1st available value selected:
image

Thus for every environment, it will get the correct kiam namespace itself.
It will even work for multiple kiams/namespaces, but can't imagine a need for this and it's not my original intention. This is only to de-hardcode namespace and make it smart enough to auto-select correct value.

### DIFF
--- a/docs/dashboard-prom.json
+++ b/docs/dashboard-prom.json
@@ -1224,7 +1224,7 @@
           "tableColumn": "",
           "targets": [
             {
-              "expr": "min(kube_daemonset_status_number_available{namespace=\"kube-system\",daemonset=~\".*kiam-server\"}) without (instance, pod)",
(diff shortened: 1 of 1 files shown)


max-lobur:
This technique is used in upstream dashboards e.g. https://github.com/helm/charts/blob/master/stable/prometheus-operator/dashboards/grafana-dashboard-k8s-resources-namespace.json#L785

Joseph-Irving:
lgtm, seems your other change has conflicted with this one, do you mind resolving the conflicts and then I can merge?

max-lobur:
@Joseph-Irving done!

PR #205, opened 2019-01-04
Disable gRPC service configuration TXT DNS record lookup in kiam agent

### LINKED ISSUE #204: Agent's gRPC client should disable TXT DNS record lookup for nonexistent gRPC service configuration
When _kiam_'s agent uses its gRPC client, it issues TXT DNS queries against the in-cluster DNS pods (CoreDNS in my case), repeatedly looking up these records to find service configuration. CoreDNS does not serve these records, and the lookups prompt log messages complaining about the client's mistaken assumptions (fixed recently in coredns/coredns#2291, addressing coredns/coredns#2290).

The Go gRPC client library used by _kiam_ gained a new option in grpc/grpc-go#2010, addressing grpc/grpc-go#1998: [`grpc.WithDisableServiceConfig`](https://github.com/grpc/grpc-go/blob/5da252b6a6c870fee942a3b18a1bc46c684f9adc/dialoptions.go#L432-L439).

At present, _kiam_ [does not use this option](https://sourcegraph.com/github.com/uswitch/kiam@4bee06120dfba42a86a14aedec0855f18d1aeac5/-/blob/pkg/server/gateway.go#L85-92) in `server.NetGateway`. Are there any objections to adding it there to disable these fruitless TXT record queries?

seh:
The _kiam_ server doesn't publish a gRPC service configuration in DNS, and the CoreDNS server used commonly in Kubernetes [doesn't serve these records anyway](https://github.com/coredns/coredns/blob/master/plugin/kubernetes/kubernetes.go#L97-98) (previously mentioned in coredns/coredns#2290), so disable this behavior in the _kiam_ agent's gRPC client.

Fixes #204.

### DIFF
--- a/pkg/server/gateway.go
+++ b/pkg/server/gateway.go
@@ -86,6 +86,7 @@ func NewGateway(ctx context.Context, address string, caFile, certificateFile, ke
 		grpc.WithTransportCredentials(creds),
 		grpc.WithUnaryInterceptor(grpc_middleware.ChainUnaryClient(retry.UnaryClientInterceptor(callOpts...), grpc_prometheus.UnaryClientInterceptor)),
 		grpc.WithBalancerName(roundrobin.Name),
+		grpc.WithDisableServiceConfig(),
 		grpc.WithBlock(),
 		grpc.WithWaitForHandshake(),
 		grpc.WithStreamInterceptor(grpc_prometheus.StreamClientInterceptor),


rverma-nikiai:
@Joseph-Irving Is this being released in any docker image yet? Thanks.

Joseph-Irving:
@rverma-nikiai this is available off of master, quay.io/uswitch/kiam:master it's not been put in a tagged  image yet

PR #209, opened 2019-01-10
Add document for prometheus setup

thejasbabu:
Documentation for kiam prometheus set-up.

### DIFF
--- a/docs/PROMETHEUS.md
+++ b/docs/PROMETHEUS.md
@@ -1,0 +1,91 @@
+# Kiam Prometheus Setup
+
+Kiam server and agent exposes prometheus metrics at `/metrics` endpoint in the default `9620` port.
+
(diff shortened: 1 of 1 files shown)


pingles:
Thanks for adding some documentation for those using the Prometheus operator. I'm not sure we'd expect all users to use the operator so maybe it's better to expand the documentation to cover Prometheus and the Prometheus Operator?

If we do that then updating the main deployment files with the prometheus metrics port is probably worth doing too (rather than relying on users to override?). Currently the manifests include annotations to scrape with vanilla Prometheus.

Anyone else have some strong opinions? @uswitch/cloud

Joseph-Irving:
Yeah I agree with @pingles, this docs is definitely useful but it would be nice to make it clear that this is only for people using the coreos prometheus operator

zytek:
This is obsolete with current built in support for service monitors

PR #213, opened 2019-02-04
Update server.yaml to allow master taints

efranford:
The kiam server needs to be schedulable on tainted nodes using a toleration for node-role.kubernetes.io/master

### DIFF
--- a/deploy/server.yaml
+++ b/deploy/server.yaml
@@ -17,6 +17,10 @@ spec:
       serviceAccountName: kiam-server
       nodeSelector:
         kubernetes.io/role: master
+      tolerations:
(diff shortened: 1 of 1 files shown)


integrii:
I don't know how this disappeared from the deployment sample yaml, but I have this in my cluster and it definitely needs added...

Joseph-Irving:
It might be worth adding a comment explaining that this is primarily for Kops or any other installations that taint the masters.

integrii:
@efranford should add a comment so this can get in there!

Joseph-Irving:
superseded by https://github.com/uswitch/kiam/pull/252

PR #215, opened 2019-02-04
add logging level to server.yaml

integrii:
A simple flag for debug level in the example will help when admins want to turn the logging up later.

### DIFF
--- a/deploy/server.yaml
+++ b/deploy/server.yaml
@@ -35,6 +35,7 @@ spec:
           args:
             - server
             - --json-log
+            - --level=info
             - --bind=0.0.0.0:443
             - --cert=/etc/kiam/tls/server.pem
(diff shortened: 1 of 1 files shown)


pingles:
@integrii thanks- is this just to highlight that the server includes a flag for controlling the log level?

integrii:
That's exactly what it is for.  It would have helped me recently so I wanted to contribute the improvement.

pingles:
Ah cool. Well I'd prefer the example config to trend towards what you'd run in production- so I'd prefer if the default level was warning please (and hopefully accomplishes the same?).

pingles:
And thanks for making a PR for something that would've helped you! Helps everyone too, love it.

integrii:
switched to warn!

PR #229, opened 2019-03-09
Implement regional STS endpoint support

### LINKED ISSUE #190: Regional STS Endpoint
Does kiam support setting a regional STS endpoint?

It looks like kube2iam supported it and we'd need to be able to set it to be able to use the STS VPC Endpoint.

cjbradfield:
fixes #190

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,7 @@ We have a [#kiam Slack channel](https://kubernetes.slack.com/messages/CBQLKVABH/
 * [Prometheus and StatsD metrics](docs/METRICS.md)
 * Uses the Kubernetes Events API to record IAM errors against the Pod so that cluster users can more readily diagnose IAM problems (via `kubectl describe pod ...`)
 * Text and JSON log formats
+* Optional regional STS endpoint support
 
 ## Overview
 From the [AWS documentation on IAM roles](http://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles.html):
--- a/cmd/kiam/server.go
+++ b/cmd/kiam/server.go
@@ -57,6 +57,7 @@ func (o *serverOptions) bind(parser parser) {
 	parser.Flag("session-duration", "Requested session duration for STS Tokens.").Default("15m").DurationVar(&o.SessionDuration)
 	parser.Flag("session-refresh", "How soon STS Tokens should be refreshed before their expiration.").Default("5m").DurationVar(&o.SessionRefresh)
 	parser.Flag("assume-role-arn", "IAM Role to assume before processing requests").Default("").StringVar(&o.AssumeRoleArn)
+	parser.Flag("region", "AWS Region to use for STS calls").Default("").StringVar(&o.Region)
 }
 
 func (opts *serverCommand) Run() {
--- a/cmd/kiam/server.go
+++ b/cmd/kiam/server.go
@@ -57,7 +57,7 @@ func (o *serverOptions) bind(parser parser) {
 	parser.Flag("session-duration", "Requested session duration for STS Tokens.").Default("15m").DurationVar(&o.SessionDuration)
 	parser.Flag("session-refresh", "How soon STS Tokens should be refreshed before their expiration.").Default("5m").DurationVar(&o.SessionRefresh)
 	parser.Flag("assume-role-arn", "IAM Role to assume before processing requests").Default("").StringVar(&o.AssumeRoleArn)
-	parser.Flag("region", "AWS Region to use for STS calls").Default("").StringVar(&o.Region)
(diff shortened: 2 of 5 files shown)


pingles:
Love it :-)

Only comment would be to maybe add a default/enum to make it easier for operators to understand what value to set in the flag?

zytek:
Will this support aws china regions? I have problem deploying `kiam` on kops-provisioned cluster in aws-cn, and this is possibly cause kiam tries to talk to global aws sts endpoint and not china local.

cjbradfield:
> Love it :-)
> 
> Only comment would be to maybe add a default/enum to make it easier for operators to understand what value to set in the flag?

Sure, I can do that. I'm thinking something like "global-endpoint" as it would be highly unlikely for AWS to use that as a region.

cjbradfield:
> Will this support aws china regions? I have problem deploying `kiam` on kops-provisioned cluster in aws-cn, and this is possibly cause kiam tries to talk to global aws sts endpoint and not china local.

The diff does support regional endpoints in China assuming they exist (I haven't tested there).

cjbradfield:
> Love it :-)
> 
> Only comment would be to maybe add a default/enum to make it easier for operators to understand what value to set in the flag?

Oh wait did you mean replicate all the region list as an enum? Although I see benefit in that, we'd have to constantly edit the list when AWS adds regions which seems error prone. Perhaps I can just add an example to the flag description?

I'm also not a fan of the previous diff I just posted (i'd prefer to leave it the zero value of the string and explain the behavior in the description). I think I misunderstood your feedback.

Anyway, thanks for the opportunity to contribute!

pingles:
Good point, don't want to needlessly replicate the list.

Maybe you could extend the description with a couple of examples so people
can see what kinds of values ought to be accepted?

Additionally, it'd be good to make sure the server exits with a failure if
the user has provided an invalid region for the host- maybe resolve the
address and if it fails to resolve successfully the server process exits
with a failure log message?

zytek:
Could region be auto-discovered from EC2 Metadata / kubernetes labels?

You could also update docs/readme to point out this new feature. 

Gonna check this in `aws-cn` today.

zytek:
I can confirm that `kiam` built from this PR works fine in AWS China. Region autodiscovery would be a nice bonus.

cjbradfield:
> I can confirm that `kiam` built from this PR works fine in AWS China. Region autodiscovery would be a nice bonus.

Hmm. Well, this would get me what I want as-is. I'll add it if everyone agrees it makes the whole thing a lot better.

The only downside I could see is it binds kiam-server to running in-cluster. That may be the case already.

zytek:
> The only downside I could see is it binds kiam-server to running in-cluster. That may be the case already. I also don't know how we would support the fips regions as well (technically, they aren't regions but the Go SDK treats them that way).

I don't think this is a big issue, as when kiam fails to discover region via EC2 Metadata it can fall back to old behavior. Anyway, this can be a scope of another PR. 

Thanks for this submission, it saved my day ;-)

pingles:
It'd be cool to have it auto-discover but I'm also happy for it to default to current behaviour (and add a second for the auto-discovery later). 

Only comment would be if we can add some help docs to the flags to indicate the _kinds_ of values you could set: `eu-west1`, `us-east1` etc. It doesn't have to be exhaustive but useful to give operators an easy way to see the kinds of values they could set.

As I said earlier, other thing I'd like (if possible) would be something that could validate and fail early if someone provides a bad region (typo etc.) as we wouldn't want to start the server, it accept connections and never be able to generate credentials.

cjbradfield:
> It'd be cool to have it auto-discover but I'm also happy for it to default to current behaviour (and add a second for the auto-discovery later).
> 
> Only comment would be if we can add some help docs to the flags to indicate the _kinds_ of values you could set: `eu-west1`, `us-east1` etc. It doesn't have to be exhaustive but useful to give operators an easy way to see the kinds of values they could set.
> 
> As I said earlier, other thing I'd like (if possible) would be something that could validate and fail early if someone provides a bad region (typo etc.) as we wouldn't want to start the server, it accept connections and never be able to generate credentials.

@pingles did you have a chance to look at the updated PR? In it I updated the flag description to include an example and also information about the default behavior.

The regional endpoint is now validated based on the existence of a region as well as DNS resolution of the hostname of the endpoint. In addition, I cache the ResolvedEndpoint so that these checks aren't performed after the initial endpoint resolution.

I have also added a bullet point to the features list in the README.md

Let me know if you want more examples. The tests also will test the resolution of the endpoint. It might be best to make one more change there to make the resolver parametric for testability so that you don't need network connectivity to run the tests.

pingles:
Fantastic, thanks for the contribution!

@Joseph-Irving be great if you guys could cut a release when you can 😄

Issue #230, opened 2019-03-11
Consider using Helm Hub to selfhost Kiam Helm chart


danmx:
Pushing changes to kiam helm chart in official helm repo to match current deployment suggestions (best practices) takes a lot of time, e.g. https://github.com/helm/charts/pull/9262.

Even maintainers suggest using [Helm Hub](https://hub.helm.sh/) so the chart could be stored in e.g. this repo.

Joseph-Irving:
Disclaimer: we don't use helm at all
However it looks like it's a bit of a pain contributing to the helm chart and I know a fair amount of people use the kiam chart so it would be good to make it a better. I think we'll have a look and see how much work it is to get it on helm hub.

danmx:
@Joseph-Irving out of curiousity what are you using instead of it and on what scale?

Joseph-Irving:
we just use kubernetes manifests deployed via our CI system, I'm sure helm is very useful in some scenarios but it has always seemed like a needless abstraction to us. Our scale is constantly changing but we've got around 600 deployments, 3k pods, three main kube clusters.

ewbankkit:
Kind of related: https://github.com/helm/charts/issues/13869.
As I mentioned [here](https://github.com/uswitch/kiam/pull/232#issuecomment-474357777) I think it would be best for others to be Owners/Maintainers on the Helm chart (wherever it's hosted) and am willing to do whatever is needed to make that happen.

maorfr:
Here is some information about hosting your own charts (copied from a similar PR):

I think that you totally should host your own charts!
This will give you more freedom around management, and you will not be blocked by us (maintainers).

This means that when users want to install your chart, instead of installing them from stable, they will install it from (for example) mycompany-charts:

$ helm install --name my-release mycompany-charts/mychart
Here are some instructions about how to host your own charts repo:
https://github.com/helm/hub/blob/master/docs/github.md#github-pages-as-a-repository

In order for people to find out about your charts repo, you should add it to helm hub:
https://hub.helm.sh/

Here are some instructions about how to add your charts repo to helm hub:
https://github.com/helm/hub#how-to-add-your-helm-charts

Hope that answered all your questions! :)

bhavin192:
> Kind of related: helm/charts#13869.
As I mentioned here I think it would be best for others to be Owners/Maintainers on the Helm chart (wherever it's hosted) and am willing to do whatever is needed to make that happen.

[...]

PR #232, opened 2019-03-15
self host helm chart

Joseph-Irving:
In relation to https://github.com/uswitch/kiam/issues/230
To make it easier for people to contribute to the kiam helm chart we can keep the definition here and self host the repo.
I've just set up an s3 bucket to use for this purpose. https://s3-eu-west-1.amazonaws.com/kiam-helm/charts


Joseph-Irving:
so it seems that it does not like being hosted in s3
```
helm repo add kiam  https://s3-eu-west-1.amazonaws.com/kiam-helm/charts
helm fetch kiam/kiam
Error: Get s3://kiam-helm/charts/kiam-2.1.0.tgz: unsupported protocol scheme "s3"
```

Strangely it works if you do a full path
```
helm fetch https://s3-eu-west-1.amazonaws.com/kiam-helm/charts/kiam-2.1.0.tgz
```

Joseph-Irving:
so that is because inside index.yaml you get
```
urls:
    - s3://kiam-helm/charts/kiam-2.1.0.tgz
```

Joseph-Irving:
seems like there is a PR on helm-s3 to fix this https://github.com/hypnoglow/helm-s3/pull/70 but it's not been merged, the alternative is to use github pages but that will require more work to implement

helm/kiam/templates/agent-daemonset.yaml
```diff
@@ -0,0 +1,132 @@
...
+        - name: tls
+          secret:
+            {{- if .Values.agent.tlsSecret }}
+            secretName: {{ .Values.agent.tlsSecret }}
+            {{else}}
+            secretName: {{ template "kiam.fullname" . }}-agent
+            {{- end }}
+        - name: xtables
+          hostPath:
+            path: /run/xtables.lock
```
danmx: Please add `type: FileOrCreate` to match https://github.com/uswitch/kiam/blob/master/deploy/agent.yaml#L32 and address https://github.com/helm/charts/issues/9261

helm/kiam/templates/agent-daemonset.yaml
```diff
@@ -0,0 +1,132 @@
...
+      {{- range .Values.agent.extraHostPathMounts }}
+        - name: {{ .name }}
+          hostPath:
+            path: {{ .hostPath }}
+      {{- end }}
+      containers:
+        - name: {{ template "kiam.name" . }}-{{ .Values.agent.name }}
+        {{- if .Values.agent.host.iptables }}
+          securityContext:
+            privileged: true
```
danmx: Please only add `NET_ADMIN` capability to match https://github.com/uswitch/kiam/blob/master/deploy/agent.yaml#L37 and improve security of the agent

helm/kiam/templates/server-daemonset.yaml
```diff
@@ -0,0 +1,145 @@
...
+{{ toYaml .Values.server.tolerations | indent 8 }}
+      {{- if .Values.server.affinity }}
+      affinity:
+{{ toYaml .Values.server.affinity | indent 10 }}
+      {{- end }}
+      volumes:
+        - name: tls
+          secret:
+            {{- if .Values.server.tlsSecret }}
+            secretName: {{ .Values.server.tlsSecret }}
```
danmx: Consider adding `optional: false`, it should prevent creation of the container when the secret is not mounted.

helm/kiam/templates/server-daemonset.yaml
```diff
@@ -0,0 +1,145 @@
...
+            {{else}}
+            secretName: {{ template "kiam.fullname" . }}-server
```
danmx: Consider adding `optional: false`, it should prevent creation of the container when the secret is not mounted.

helm/kiam/templates/agent-daemonset.yaml
```diff
@@ -0,0 +1,132 @@
...
+{{ toYaml .Values.agent.tolerations | indent 8 }}
+      {{- if .Values.agent.affinity }}
+      affinity:
+{{ toYaml .Values.agent.affinity | indent 10 }}
+      {{- end }}
+      volumes:
+        - name: tls
+          secret:
+            {{- if .Values.agent.tlsSecret }}
+            secretName: {{ .Values.agent.tlsSecret }}
```
danmx: Consider adding `optional: false`, it should prevent creation of the container when the secret is not mounted.

helm/kiam/templates/agent-daemonset.yaml
```diff
@@ -0,0 +1,132 @@
...
+            {{else}}
+            secretName: {{ template "kiam.fullname" . }}-agent
```
danmx: Consider adding `optional: false`, it should prevent creation of the container when the secret is not mounted.

ewbankkit:
I'm happy to be replaced in the _maintainers_ list in `Chart.yaml`.
Even though we're big users of `kiam` I have a myriad of other competing claims on my time 😄.

danmx:
@Joseph-Irving would you mind if I contribute my suggestions to your branch?

Joseph-Irving:
@danmx, sorry not really looked at this as the self hosting does not currently work. I'd rather get the hosting working first and then look at adding your suggestions.

danmx:
> seems like there is a PR on helm-s3 to fix this [hypnoglow/helm-s3#70](https://github.com/hypnoglow/helm-s3/pull/70) but it's not been merged, the alternative is to use github pages but that will require more work to implement

It seems that project is not maintained anymore

danmx:
why not use `helm package` to create an archive and directly upload it to S3?

PR #235, opened 2019-03-22
Sample deployment on EKS with terraform

fbreckle:


pingles:
Awesome, thanks @fbreckle!

@uswitch/cloud this would be amazing to execute in a standalone Kiam AWS account as part of the build.

Joseph-Irving:
Hey @fbreckle, sorry for taking so long to review this, I've struggled to find the time to go through it all carefully. Just want to triple check all the terraform is sensible before encouraging users to try it out.

deploy/eks-terraform/README.md
```diff
@@ -0,0 +1,17 @@
...
+This is a sample deployment of kiam on Amazon EKS with minimal dependencies.
+
+## Dependencies
+
+- terraform for creating the resources on AWS
+- aws cli
+- aws-iam-authenticator for authenticating to the cluster
+- kubectl
+- cfssl
+- basic linux tools
```
Joseph-Irving: could you add links to these tools to make it easier to install?
e.g [cfssl](https://github.com/cloudflare/cfssl).

fbreckle: Sure. aws-iam-authenticator is not even required anymore since the aws cli can do that now.
I should be able to do it this week.

deploy/eks-terraform/kiam.tf
```diff
@@ -0,0 +1,51 @@
+#
+# Create the intermediary Kiam role
+#
+resource "aws_iam_role" "kiam_intermediary" {
+  name        = "eks-${var.cluster_name}-kiam-intermediary"
+  description = "Role the Kiam server process assumes. If you want to use Kiam to assume any role R, THIS is the role that R must trust."
```
Joseph-Irving: I would maybe rephrase this slightly, something like "Role the Kiam server process assumes. Any role that you want Kiam to assume must trust **THIS** role."

Joseph-Irving:
I think this overall looks pretty good, expanding the readme to give a bit more explanation would be useful. It could do with an explanation of what it will create, e.g IAM roles, kiam server node pool, eks cluster with nodes etc and some brief reasoning alongside it. E.g we run kiam on a tainted set of node as we do not have access to that masters in EKS. 

Maybe also include a note suggesting this is just an example and not verified for production usage. Also have a link from the main readme to this page.

deploy/eks-terraform/kiam.tf
```diff
@@ -3,7 +3,7 @@
 #
 resource "aws_iam_role" "kiam_intermediary" {
   name        = "eks-${var.cluster_name}-kiam-intermediary"
-  description = "Role the Kiam server process assumes. If you want to use Kiam to assume any role R, THIS is the role that R must trust."
+  description = "Role the Kiam server process assumes. Any role that you Kiam to assume must trust THIS role."
```
Joseph-Irving: Any role that you -> Any role that you want

PR #249, opened 2019-05-07
Pod cache should only be watching running pods

surindersinghp:
We see the following error on our clusters because k8s starts reusing the ips from pods stuck in pending state: "error finding pod: multiple running pods found"

shouldn't it be looking at running pods only?

### DIFF
--- a/pkg/k8s/listwatch.go
+++ b/pkg/k8s/listwatch.go
@@ -15,5 +15,5 @@ const (
 
 // NewListWatch creates a ListWatch for the specified Resource
 func NewListWatch(client *kubernetes.Clientset, resource string) *cache.ListWatch {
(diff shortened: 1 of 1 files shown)


pingles:
It attempts to ensure that credentials are loaded and fetched as fast as possible: if only running pods were tracked then caches would never be able to prefetch. Maybe there's a way to achieve it through changing the implementation to separate the prefetching/credentials processes and the resolution of Pods from their IPs but I don't think that would guarantee it.

If AWS hadn't announced official IAM integration on their roadmap we'd probably work on extending the changes we made for v3: this sets up the ability to have agents operate as sidecars to Pod containers and would let us use service account identity (#91) to avoid needing any IP lookup at all. Maybe it's worth scoping that out in case people are interested in implementing.

pingles:
I'm going to close this for now as it's a design decision to work the way it does.

Happy for a discussion in future if we want though.

PR #250, opened 2019-05-13
Treat ErrPolicyForbidden and ErrPodNotFound as permanent errors during retry backoff

hoelzro:
If a pod requests credentials it's not allowed to, kiam-agent will keep retrying the same request until the five second timeout is reached.  This changeset treats ErrPolicyForbidden (as well as ErrPodNotFound) as permanent errors, so that the AWS client talking to kiam-agent gets a response in a more timely fashion.

### DIFF
--- a/pkg/aws/metadata/handler_credentials.go
+++ b/pkg/aws/metadata/handler_credentials.go
@@ -24,8 +24,6 @@ import (
 	"github.com/uswitch/kiam/pkg/server"
 	"github.com/uswitch/kiam/pkg/statsd"
 	"net/http"
-
-	status "google.golang.org/grpc/status"
 )
 
 type credentialsHandler struct {
@@ -77,13 +75,9 @@ func (c *credentialsHandler) fetchCredentials(ctx context.Context, ip, requested
 	op := func() error {
 		creds, err := c.client.GetCredentials(ctx, ip, requestedRole)
 		if err != nil {
-			if grpcStatus, ok := status.FromError(err); ok {
-				switch grpcStatus.Message() {
-				case server.ErrPolicyForbidden.Error():
-					return backoff.Permanent(server.ErrPolicyForbidden)
-				}
+			if err == server.ErrPolicyForbidden {
+				return backoff.Permanent(err)
 			}
-
 			return err
 		}
 		credsCh <- creds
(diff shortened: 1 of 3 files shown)


hoelzro:
On a side note, I would have preferred to avoid string comparisons for the error messages, but both of these error types are currently mapped to "Unknown" at the gRPC layer.  Changing those would be more involved and potentially disruptive to other users of KIAM, but if you think it would be worth doing, please let me know and I can submit another PR with that change as well.

pingles:
It's a 👍 from me, @uswitch/cloud?

pingles:
Actually, having just read... I'd support this change for `ErrPolicyForbidden` but not `ErrPodNotFound`.

My rationale would be that there's a chance the cache/watcher will be behind when the pod requests credentials- causing `ErrPodNotFound` to be intermittent. I think it'd be better for the agent to keep retrying up to the SDK disconnection.

Any objection @hoelzro?

hoelzro:
@pingles Yeah, that makes sense to me!  I've added a commit removing permanent error status for `ErrPodNotFound`.

pingles:
Thanks! Could you try and find a way to add a test too please- it's a little verbose but should be relatively easy to do in https://github.com/uswitch/kiam/blob/master/pkg/aws/metadata/handler_credentials_test.go

hoelzro:
Test added!  There are a few things that came to mind while writing the test:

  * I changed the code so that the gateway client translates gRPC errors into `server.ErrPolicyForbidden` and `server.ErrPodNotFound` - I felt that this belonged in the code directly invoking the RPC routines.
  * The test tests for a 500 error - there's no real specification for what the EC2 metadata server returns on error, but I figured that would be fine.  What do you think?  I happen to know that boto, for example, retries on every non-200 status, so maybe how clients handle different statuses should be the guiding star.
  * I noticed that while `GetPodCredentials` will translate a `k8s.ErrPodNotFound` into `server.ErrPodNotFound`, `GetPodRole` does not - should the code for the latter be updated to do that?

pingles:
Sorry it's taken me ages to take a look; I've been stacked at work. I'll see if I can get a chance to review next week unless one of @uswitch/cloud beat me to it 😄

hoelzro:
@pingles No worries - please ping me when you have a chance to review it!

hoelzro:
@pingles Have you or @uswitch/cloud had a chance to look at this?

pingles:
I'm really sorry I haven't. It looks ok to me, @tombooth / @Joseph-Irving could you take a look too please?

hoelzro:
@pingles @Joseph-Irving Thanks!

Issue #202, opened 2018-12-19
On-the-fly agent replacement


max-lobur:
Environment: AWS, k8s 1.10, kops, kiam helm chart from this PR https://github.com/helm/charts/pull/9262 , quay.io/uswitch/kiam:v3.0-rc1.

We've got into an issue when did kiam upgrade:
- helm upgrade
- rotate kiam server pods 1 at time
- rotate kiam agent pods 1 at time.

Server part went smooth. 
After agent part, the new agent pod became invisible for all the existing pods: their metadata calls were not intercepted by kiam agent and they were getting bare node role instead of one set in the annotation. This condition persists for at least 30 min, after that I re-create all affected pods and they started to work.

Right now we are doing node replacement to rotate an agent. Am I missing something in docs, is this expected, or can be avoided?

pingles:
Thanks for reporting the problems you had. The agent command has a flag that will have it [insert an iptables rule](https://github.com/uswitch/kiam/blob/master/cmd/kiam/iptables.go) to intercept metadata traffic and it also removes it by default when the process terminates (which could leave a window where the iptables rule isn't inserted).

One option is to change your node configuration to insert the iptables rule as part of the system initialisation process. I'd also happily take a PR for an additional agent flag for something like `--no-iptables-removal-on-termination` to prevent Kiam from dropping the rule on termination.

As for pods failing for longer, that _feels_ like an issue [we've seen in the past with poor application behaviour](https://medium.com/@pingles/kiam-iterating-for-security-and-reliability-5e793ab93ec3): the standard AWS SDKs appear to cause exceptions/errors as soon as they're unable to access the metadata API (or the kiam agent) even if they have credentials that remain valid. Not all applications have failure handling in the right places (so a call to send an SQS message could fail because the credentials are no longer valid but the application won't recreate an authenticated session, for example). Could you confirm whether this is a plausible explanation for what you experienced? 

Based on what you describe it sounds like setting iptables rules on node boot/preventing kiam from deleting would help although probably won't guarantee you wouldn't see the problem again. Node replacement will guarantee you won't see the problem (and doesn't require application changes), alternatively I think it's probably a case of reviewing the failure handling within applications.

max-lobur:
@pingles ++ on system level rule installation. Will implement that.

This is not an app issue for sure. We used those apps for half a year already on kiam v2, with zero issues. I got into this once upgraded kiam to v3-rc1 and replaced the agent on a running node.

Steps to reproduce:
[...]

PR #253, opened 2019-06-05
Allow kiam-agent to not remove rules on the host

theatrus:
There are cases where kiam should not remove the iptables rules at
shutdown. This avoids cases where metadata from the host will leak
when kiam is reprovisioning, has crashed, etc, which causes any AWS
SDK to cache and try to use host credentials for hours on end.


Nuru:
Does this fix #202?

theatrus:
Yes, it looks like it would fix that issue.

PR #255, opened 2019-06-13
Support External ID

### LINKED ISSUE #236: External_ID Support for Cross account IAM Role
AWS IAM supports external id as a security measure for cross account IAM roles.
Ref: https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

Issue for enhancement to add support for external id in KIAM.

randomvariable:
Fixes #236 

This supports External ID via annotations as per updated [readme](https://github.com/uswitch/kiam/blob/97c76c3e7e4e300b92de6016a39aa52d3f878cf3/README.md#overview).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ bin/
 kiam.test
 mem.out
 coverage.txt
+.env
--- a/README.md
+++ b/README.md
@@ -48,6 +48,20 @@ metadata:
     iam.amazonaws.com/role: reportingdb-reader
 ```
 
+kiam also supports the use of an external id annotation added to the `Pod` which
+maybe used to avoid [confused deputy scenarios in cross-organisation role assumption][confused-deputy],
+for example:
+
+```yaml
+kind: Pod
+metadata:
+  name: foo
+  namespace: cross-org-iam-example
+  annotations:
+    iam.amazonaws.com/role: arn:aws:iam::12345678901:role/trusted-partner
+    iam.amazonaws.com/external-id: d272840b-2859-41aa-b022-f4527bfef672
+```
+
 Further, all namespaces must also have an annotation with a regular expression expressing which roles are permitted to be assumed within that namespace. **Without the namespace annotation the pod will be unable to assume any roles.**
 
 ```yaml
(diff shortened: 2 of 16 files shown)


pingles:
Exciting, thanks for working on this @randomvariable

harishspqr:
@pingles Could you please share whats blocking this PR and why the checks have failed ? I am unable to open the details about this failure.

randomvariable:
@harishspqr I hadn't finished it previously, but have done now via the annotations.

Have had a look to see if we could support external ID via a reference to a secret or some other thing.

One of the issues is that if we leveraged a SecretRef, kiam's server would need RBAC permissions to potentially read all secrets in all namespaces, which is less than ideal.

Could get around this by introducing a new CRD, as in this branch: https://github.com/randomvariable/kiam/tree/external-id-crd
([definition](https://github.com/randomvariable/kiam/blob/external-id-crd/config/crds/iam_v1alpha1_iamrole.yaml))

Then kiam can have an RBAC policy to be able to inspect only those CRD types.

The approach in the branch is pretty much that of PodSecurityPolicies - read all the IAM CRDs, find the service account of the pod, and find the first IAM object that passes SubjectAccessReview when ordered alphabetically. To finish that branch, someone would need to continue to rework the interfaces and the cache behaviour to make it work. Potentially a big enough change it should be in a different repository.

That said, this PR is ready in terms of a simple extension via an annotation.

harishspqr:
@randomvariable Awesome! 
Thanks for the details.

harishspqr:
@randomvariable Can we create a secret in a particular namespace and create Role that is limited to the secrets resource in that namespace alone ? Is it possible ?

randomvariable:
@harishspqr Maybe, if it's done as a SecretReference. I probably won't have any time to do anything like that, but if you want to run with it in a separate PR, feel free.

danmx:
@randomvariable are you proposing moving `kiam` towards an operator?

randomvariable:
@pingles @Joseph-Irving I've rebased this on master. Ready for merge.

randomvariable:
Replaced by https://github.com/uswitch/kiam/pull/360

PR #263, opened 2019-07-08
Add support foriam.amazonaws.com/allowed-roles namespace annotation

### LINKED ISSUE #26: Divergent annotations for kiam/kube2iam projects, an end user plea
Hi all and thanks for your efforts on `kiam`, I really like the idea of pre-caching credentials for faster container starts.

As an end user it would be fantastic if the `kaim` and `kube2iam` implementations could agree to support that same annotations. This make it easier for `kube2iam` users to switch to `kiam` or to try it and switch back. Or to deploy applications to two clusters, one using each implementation.

Right now the Pod annotation is the same for both - **fantastic!**
```
kind: Pod
metadata:
  name: foo
  namespace: iam-example
  annotations:
    iam.amazonaws.com/role: reportingdb-reader
```

But the namespace annotation has diverged. `kiam` using a single regex to match roles:
```
kind: Namespace
metadata:
  name: iam-example
  annotations:
    iam.amazonaws.com/permitted: ".*"
```
And `kubei2iam` uses a list of roles or role wildcards:
```
Kind: Namespace
metadata:
  name: iam-example
  annotations:
    iam.amazonaws.com/allowed-roles:
    - "role-arn"
    - "my-custom-path/*"
    - "*"
```
Now I am sure both projects had their reasons and preferences. But again, as an end users of both these great projects, I look at the divergent namespace annotations and say "Really!??" 😄 

IMHO, it doesn't seem like either approach is significantly better in any practical way, or in a way that makes this divergence beneficial to users. I feel the use cases for compatibility are strong, and I don't perceive what the use cases are for this divergence?

**Pretty please, could the projects not make friends, get drunk together if necessary, and pick either or both namespace annotation options, and both projects support the same one(s)?** 🍻 

Or second best, if you can't agree, could `kiam` also support `iam.amazonaws.com/allowed-roles` by compiling the list of `allowed-roles` roles into a regex to use internally?

jacksontj:
This adds support for an additional namespace annotation for which roles are allowed for a given pod

This is a first-pass, as it was a bit unclear from the issue what the desired end-state should be. This PR right now adds support for glob annotations on `iam.amazonaws.com/allowed-roles` (how kube2iam works) -- but without support for regex in that annotation.

I'm assuming that the end-state we want is a single annotation (`iam.amazonaws.com/allowed-roles` -- which would presumably under-the-covers convert all old annotation values to the new one for compatibility), so I'm seeing options:

1. Allow both regex and glob at the same time: some checking for "is this a regex" (if we require `^$` its not beautiful, but works). 
    - pros:
         - Any regex we found in the "old" mechanism could be wrapped in those characters if they weren't there already.
    - cons:
         - Regex definitions from the kube2iam mechanism won't work without changes (although adding the anchors should work with both).
2. add a flag (similar to kube2iam) that defines whether the list in `iam.amazonaws.com/allowed-roles` is a list of globs or regexes. 
    - pros:
        - this would make it compatible (more or less) with kube2iam
    - cons:
        - the "old" annotation could only be used if the CLI flag was set to regex-- otherwise it would be interpreted as a glob
        - This makes configuration split between 2 places (cli falgs + annotations): not in favor of this approach (even though this is the kube2iam approach).configuration split between 2 places, which makes the configuration unclear
3. (which I don't think we should do, but calling it out anyways) change the list to be a list of struct instead of string -- and in there we define if its a string or a regex.
    - pros:
        - most explicit about what the matcher is
    - cons:
        - not compatible directly with *any* existing annotation
4. ? (can't think of another)

cc @pingles 
Fixes #26

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -3,20 +3,25 @@
 
 [[projects]]
   branch = "master"
+  digest = "1:315c5f2f60c76d89b871c73f9bd5fe689cad96597afd50fb9992228ef80bdd34"
   name = "github.com/alecthomas/template"
   packages = [
     ".",
-    "parse"
+    "parse",
   ]
+  pruneopts = "UT"
   revision = "a0175ee3bccc567396460bf5acd36800cb10c49c"
 
 [[projects]]
   branch = "master"
+  digest = "1:c198fdc381e898e8fb62b8eb62758195091c313ad18e52a3067366e1dda2fb3c"
   name = "github.com/alecthomas/units"
   packages = ["."]
+  pruneopts = "UT"
   revision = "2efee857e7cfd4f3d0138cc3cbb1b4966962b93a"
 
 [[projects]]
+  digest = "1:663aa72b14f781c4f3f770642fe0c6331e450f45a4c579ffd8fc35387ebb78a6"
   name = "github.com/aws/aws-sdk-go"
   packages = [
     "aws",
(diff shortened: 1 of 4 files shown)


jacksontj:
Unfortunately it seems that I can't access the output of the CI run without authorizing some github app with waaaay too many permissions :/

pingles:
The error reported in the build:

```
+ make test
go test github.com/uswitch/kiam/pkg/... -race
pkg/server/policy.go:21:2: cannot find package "github.com/ryanuber/go-glob" in any of:
	/go/src/github.com/uswitch/kiam/vendor/github.com/ryanuber/go-glob (vendor tree)
	/usr/local/go/src/github.com/ryanuber/go-glob (from $GOROOT)
	/go/src/github.com/ryanuber/go-glob (from $GOPATH)
Makefile:27: recipe for target 'test' failed
make: *** [test] Error 1
```

pingles:
I guess the issue is that detecting globs / regexes is hard as a regex can be almost anything- so globs will mostly look like regexes?

Having said that, I believe Kiam and Kube2iam use different annotations (and I have a feeling I chose deliberately to do this to avoid this kind of problem).

* Kiam: `iam.amazonaws.com/permitted`
* Kube2iam: `iam.amazonaws.com/allowed-roles`

In which case, I'd be inclined to support either annotation: globs for the latter, regexes for the former. There oughtn't be any issue with supporting two but perhaps adding a flag on the server to add support for `allowed-roles` would be useful for cluster operators to encourage consistency where possible.

jacksontj:
@pingles personally I prefer the list/glob as its much more explicit; So maybe we should add some sort of flag to specify which annotation to look at, I can look into adding that as well.

As for the CI runs, is there a way to get that output without granting all those perms (seemed like a lot :scream: ), without looking at that info its going to take a fair amount of back and forth.

pingles:
Alternatively, we could offer two flags: `--permitted-roles-key` to set the annotation name (defaults to `iam.amazonaws.com/permitted` to maintain current behaviour. Second flag is `--permitted-roles-type` is an enum of either `regex`, or `glob` to switch between behaviours.

That'd allow operators to maintain kube2iam compatibility or switch between globs and regexes separately.

As for the build- what permissions does Drone request?

jacksontj:
In that flag-controlled model we couldn't support both for a time (which is probably okay?) and that would mean its either a list of globs or a single regex (whereas kube2iam supports lists of regexes).


As for permissions:

```
This application will be able to read and write all public and private repository data. This includes the following:

    Code
    Issues
    Pull requests
    Wikis
    Settings
    Webhooks and services
    Deploy keys
    Collaboration invites
```

The rest is fine, but that seems wildly intrusive

jacksontj:
Summarizing the conversation we had on k8s slack:

* we want to support both regex and glob
* we want to enable migrations between kube2iam and kiam
* we'll allow for both options to be enabled/disabled separately
* we'll allow user to define the annotation name for both

flags will look something like `roles-annotation-regex` (define the annotation for the single regex -- same behavior as today) and `roles-annotation-globs` (define annotation for the list of globs -- default kube2iam behavior).

PR #264, opened 2019-07-09
Update Dockerfile to include ca-certificates

taharah:
* Update to Golang v1.15.7 base image
* Update final base image to Alpine v3.13
* Add ca-certificates to final build image. Required for communicating with the AWS API.
* Formatting changes. 

Signed-off-by: Trevor Wood <Trevor.G.Wood@gmail.com>

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -5,7 +5,7 @@ workspace:
 pipeline:
   test:
     secrets: [ codecov_token ]
-    image: golang:1.13
+    image: golang:1.15
     environment:
       GO111MODULE: "on"
     commands:
@@ -14,7 +14,7 @@ pipeline:
       - ./ci/codecov.sh
 
   benchmarks:
-    image: golang:1.13
+    image: golang:1.15
     environment:
       GO111MODULE: "on"
     commands:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.15.7-buster as build
+FROM golang:1.15.7 as build
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,4 @@
-FROM golang:1.13.8 as build
-ENV GO111MODULE=on
+FROM golang:1.15.7-buster as build
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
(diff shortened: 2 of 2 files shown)


pingles:
Thanks for your PR.

Updating the version of Golang, Alpine is cool. I'm on-the-fence about running tests during the image build, anyone else have strong opinions?

However, I do prefer the certificates out of the image so they can be managed/updated on the hosts of the cluster (without needing us to build new images of each container). Again, it's come up a few times so I'm more in favour of keeping as-we-are but I'm open to a discussion.

taharah:
I can remove the tests from the image build. If the tests are already ran as part of the CI pipeline for releases, then it would be redundant; however, it would be useful for local development and testing.

Additionally, it seems to be fairly standard to include the public CA certificates. The public CA bundle does not change that often, and for cases where users still need something different, e.g., to add custom CAs, then this does not prevent them from still being able to bind mount the CA bundle from the host. However, I can also remove this change if that is what is desired.

hairyhenderson:
I know this is an old PR, but is there any chance of getting this (or an updated version of this) merged? 

We were affected by a recent change in the helm chart, around how the certificates are mounted in. This is unusual and we would much rather have the certs shipped in the image (just like 99% of other images that do TLS things do).

hairyhenderson:
Also, aside from including certs, it would be good to just get the version of Alpine (and Go) updated - currently image scans flag this image as having a number of CVEs against it:

- CVE-2019-14697 (in `musl`)
- CVE-2018-20679 (in `busybox`)
- CVE-2019-5747 (in `busybox`)

I believe that simply updating to a recent Alpine image (like `alpine:3.11`) would suffice!

Thanks.

/cc @taharah @pingles

hfuss:
We run kiam outside of AWS, and provide IAM user creds directly to the server component. As a result, we do not need the underlying hosts' certs in order to trust the metadata API. We'd prefer to avoid the host mount altogether in this case. So in addition to including the certs in the container, it would be great if the hostPath mount was optional for the `server` component in the chart.

Before we tackle the chart, would a compromise be either producing a secondary image or tag with the CA certs included?

I know it wouldn't be great to have to maintain the two, but that way by default it still ships without the certs, remaining backwards compat in that sense, and then users can opt-in to having the CA certs by overriding the image/tag.

PR #268, opened 2019-07-22
Add deeper healthcheck for agent

mattmb:
Currently the agent health endpoint just checks that it can return some
metadata from AWS. I would also like to be able to check that the agent
is able to communticate succsessfully with the server. The motivation is
that we are renewing certificates externally and when the client cert
expires the agent will fail to talk to the server. In this case I want
the healthcheck to fail so that kubernetes will restart the agent and
cause it to re-read (the now renewed certs) from disk.

I've implemented this as a URL query param so that doing `/health` will
continue to function the same but `/health?deep=anything` will also do a
grpc call to the server and assert the health endpoint there returns
"ok"

### DIFF
--- a/pkg/aws/metadata/handler_health.go
+++ b/pkg/aws/metadata/handler_health.go
@@ -16,14 +16,17 @@ package metadata
 import (
 	"context"
 	"fmt"
+	"github.com/cenkalti/backoff"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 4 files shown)


dbyron0:
Thanks for this!

mattmb:
@pingles any thoughts on this?

pingles:
Looks good, @uswitch/cloud comments?

PR #269, opened 2019-07-22
[Grafana dashboard] Use correct label matching operators

bhavin192:
- Fixes some of the places where regex was used but the matching
  operator was `=` instead of `=~`
- Reference: https://prometheus.io/docs/prometheus/latest/querying/basics/#time-series-selectors


Joseph-Irving:
thanks!

PR #279, opened 2019-08-14
Update grpc-go from 1.14.0 to 1.23.0

DReigada:
The grpc-go version currently used is a bit outdated (1 year old), updating the version will bring some stability improvements as well as bug fixes.


pkg/server/gateway.go
```diff
@@ -90,7 +90,6 @@ func NewGateway(ctx context.Context, address string, caFile, certificateFile, ke
 		grpc.WithBalancerName(roundrobin.Name),
 		grpc.WithDisableServiceConfig(),
 		grpc.WithBlock(),
-		grpc.WithWaitForHandshake(),
```
DReigada: This is already enabled by default (the method was also removed)

pingles:
Thanks @DReigada. Funnily enough we were talking about this upgrade yesterday! @uswitch/cloud should pick this up.

DReigada:
Thanks for the quick feedback and approval! 👍

PR #276, opened 2019-07-25
Use resolver to create prefix for assumeRoleArn

derrickburns:
Addresses #271

PR #280, opened 2019-08-15
return empty string if resolver passed empty role

Joseph-Irving:
https://github.com/uswitch/kiam/pull/276 introduced a bug if you don't set assume-role-arn

`arnResolver.Resolve(config.AssumeRoleArn)` returns something like `arn:aws:iam::xxxxxxxxx:role/` when passed an empty assume role arn, which prevents the Default Gateway from using the instance role, it instead trys to assume this non existent role.

So I've changed `arnResolver.Resolve` to return "" if the role is empty, restoring the old behaviour

### DIFF
--- a/pkg/aws/sts/arn_resolver.go
+++ b/pkg/aws/sts/arn_resolver.go
@@ -30,6 +30,10 @@ func DefaultResolver(prefix string) *Resolver {
 
 // Resolve converts from a role string into the absolute role arn.
 func (r *Resolver) Resolve(role string) string {
+	if role == "" {
+		return ""
+	}
+
 	if strings.HasPrefix(role, "/") {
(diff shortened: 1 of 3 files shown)


pkg/aws/sts/arn_resolver.go
```diff
@@ -30,6 +30,11 @@ func DefaultResolver(prefix string) *Resolver {
 
 // Resolve converts from a role string into the absolute role arn.
 func (r *Resolver) Resolve(role string) string {
+
```
pingles: @Joseph-Irving remove this whitespace line?

Joseph-Irving: yep done

PR #283, opened 2019-08-20
Create self hosted helm charts

rhysemmas:
This PR includes the files needed to create packaged Helm charts which we can host ourselves. It also includes another CI step to package the Helm charts and update the repository where they are hosted.


rhysemmas:
Relates to #230

Issue #188, opened 2018-11-28
Create flag to default role whitelist regex - support drop-in kube2iam replacement


nhoughto:
So we have `kube2iam` to `kiam` migration on the backlog, but the extra requirement to annotate each namespace that is going to assume-role from is extra work for us to migrate from kube2iam to kiam. I understand why this configuration exists, and why it is the default, but it would be useful to be able to set a default allowed role regex as a flag like kube2iam does to simplify the migration path from one to the other. That way we can get the benefits, and tighten the screws around who can assume what,, when later rather than having to do both upfront.

Thanks!

bkruger99:
This!   If you also have a larger installation where you have products that use metadata, this is also a blocker.  Tagging the namespaces is easy enough, but the pods to default to the node credentials until they can be tightened down would be extremely helpful.

PR #290, opened 2019-09-04
Enable passing of default role-filtering expression.

Milesjpool:
A key step and inhibitor in migrating from Kube2IAM to Kiam is adding role-filtering patterns to each namespace of interest - e.g. #188. 

This change adds an option to enable the supplying of a default pattern, for use when the pattern is not defined in any given namespace.

The option has been named _'role-pattern-default'_, but feel free to request changes to this.

### DIFF
--- a/cmd/kiam/server.go
+++ b/cmd/kiam/server.go
@@ -58,7 +58,7 @@ func (o *serverOptions) bind(parser parser) {
 	parser.Flag("session-refresh", "How soon STS Tokens should be refreshed before their expiration.").Default("5m").DurationVar(&o.SessionRefresh)
 	parser.Flag("assume-role-arn", "IAM Role to assume before processing requests").Default("").StringVar(&o.AssumeRoleArn)
 	parser.Flag("region", "AWS Region to use for regional STS calls (e.g. us-west-2). Defaults to the global endpoint.").Default("").StringVar(&o.Region)
-	parser.Flag("allow-all-namespace-roles", "").BoolVar(&o.AnyNamespaceRole)
+	parser.Flag("role-pattern-default", "Role filtering pattern used when not defined in namespace").Default("").StringVar(&o.RolePatternDefault)
 }
 
 func (opts *serverCommand) Run() {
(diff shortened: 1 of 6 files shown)


pingles:
Thanks for your contribution @Milesjpool 

I'd make a small suggestion that the flag be named with `namespace` to help know it's the namespace policy expression (`namespace-role-pattern-default` perhaps?).

In the past I've been hesitant to add support for defaults within Kiam (default roles for pods that don't have an annotation, or default expressions in namespaces). I think there's probably 2 reasons:

1) I like the idea of encouraging the behaviours we want (i.e. seeing explicit labeling of responsibilities for pods and namespaces)
2) It could introduce subtler bugs. It's less of an issue in this situation, but, because the pod and namespace data is delivered via a Kubernetes API watcher there's a chance that a Pod will request credentials ahead of the state having been tracked by Kiam; in such a situation the default value would be used rather than the explicitly labeled (or not) value. Although applications should manage such errors it could be a pretty subtle bug to diagnose.

Having said that, I'm all for helping cluster operators do their job so I'd love to hear what other people in the Kiam community think. I think adding it as an optional flag is a sound idea giving operators the opportunity to decide but would be interested to hear what people thought on the issue before merging.

Thanks again for taking the time to contribute- we really appreciate it!

Milesjpool:
Thanks for the feedback, and the suggestion - I'll change the flag name and leave this open for further comment.

I do recognise your first point re. encouraging good-practice. For our use case this change is an enabler for an incremental migration; in the long term we would aim to annotate each namespace correctly. Perhaps we can consider strategies to discourage long-term use of this in production?

nhoughto:
Can this be merged? Defaults are are a _super_ helpful tool to aid migration, that can then be turned off later.

pingles:
I'm happy for this to be merged (although the other v4 work means there'll need to be some work to bring up to date).

I wonder if there's a way to better record and publish the events on these things though and make it easier to know when a pod is obtaining credentials using default policies. I think I'd prefer we make it very obvious in any situation we lean on a default, as opposed to what's been explicitly annotated.

PR #292, opened 2019-09-04
Change Helm extraEnv parameters to array

### LINKED ISSUE #286: Changelog for helm/stable/kiam
As I'm using helm/kiam and upgrading from 2.1.0 it would be good to know what changed.

Is there a changelog for helm/stable/kiam? I generated one from Commits but it only include 2.5.1. 

Is there anything prior to this?

Very short generated CHANGELOG.md that could be a good start?

```
##   2.5.1

Updated READMEs and include uSwitch logo in chart
commit: 9c2bc7a 

##   2.5.1

Update maintainer
commit: 39d5ef5 

##   2.5.1

Add helm chart files and ci step
commit: cc349ca 
```

### LINKED ISSUE #289: Helm: support passing env from configmap/secrets
Right now, the existing extraEnv option that can be passed is rather limiting since it only supports key/value pairs. When you want to store information like the ARN of the role to be assumed by the server process, it'd be nice to be able to reference that from a configmap or secret since it's possible that it's generated by something like terraform, so it's not something that can be known ahead of time.

My proposal is to change it to support an array of options, similar to the [external-dns chart](https://github.com/helm/charts/blob/7ac49621ca49418f23655dba856a4f5a698e84d1/stable/external-dns/values.yaml#L250-L258). For each object in that array, you could pass a direct value, or use configMapRef or secretKeyRef.

This would be a breaking change to values.yaml and would require a major version bump.

rhysemmas:
* Fix #289 - Update extraEnv parameters to arrays, allowing env vars to be created from configmaps and secrets, as well as key:value pairs
* Fix #286 - Create changelog to track Helm chart changes

### DIFF
--- a/helm/kiam/CHANGELOG.md
+++ b/helm/kiam/CHANGELOG.md
@@ -1,0 +1,27 @@
+# Helm Chart Changelog
+## 3.0.0
+4 September 2019
+
+Notable Changes:
+* The `extraEnv` parameters in `values.yaml` have been changed to support an array of options.
+
+## 2.5.3
+29 August 2019
+
+Notable Changes:
+* [#288](https://github.com/uswitch/kiam/pull/288) Bug fix for correctly rendering port number for certificates.
+
+Many thanks to the following contributor for this release:
+* [@simnalamburt](https://github.com/simnalamburt)
+
+## 2.5.2
+27 August 2019
+
+Notable Changes:
+* [#285](https://github.com/uswitch/kiam/pull/285) and [#287](https://github.com/uswitch/kiam/pull/287) Chart is updated to include uSwitch logo.
+
+## 2.5.1
+20 August 2019
+
+Notable Changes:
+* [#283](https://github.com/uswitch/kiam/pull/283) Kiam Helm charts are added to the [uswitch/kiam](https://github.com/uswitch/kiam) repo.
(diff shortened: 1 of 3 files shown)


helm/kiam/CHANGELOG.md
```diff
@@ -0,0 +1,27 @@
+# Helm Chart Changelog
+## 3.0.0
+4 September 2019
+
+Notable Changes:
+* [#292](https://github.com/uswitch/kiam/pull/292) The `extraEnv` parameters in `values.yaml` have been changed to support an array of options.
```
Joseph-Irving: I think you should make clear that this is a breaking change and what you need to do to fix your stuff if you are using extraEnv

helm/kiam/CHANGELOG.md
```diff
@@ -0,0 +1,28 @@
+# Helm Chart Changelog
+## 3.0.0
+4 September 2019
+
+Notable Changes:
+* [#292](https://github.com/uswitch/kiam/pull/292) The `extraEnv` parameters for both the agent and server in `values.yaml` have been changed to support an array of options.
+This will break any existing Helm deployments which utilise the `extraEnv` parameters in `values.yaml`. You will need to update your `values.yaml` file to match the format in the [template](/helm/kiam/values.yaml#L93)
```
Joseph-Irving: I was envisioning BIG BOLD TEXT and alarms

PR #294, opened 2019-09-09
Set chart gatewayTimeoutCreation to project deployment defaults

raehik:
Migrated from [helm/charts!16369](https://github.com/helm/charts/pull/16369).

The Helm chart default of 50ms can cause pods to restart randomly on
start-up, exacerbated by any CPU throttling or resource limits. The
[default set in the project manifests](https://github.com/uswitch/kiam/blob/04e66a7203457ada3b2dfb2efd0e9fdc50ea9260/deploy/agent.yaml#L56) is 1s.

### DIFF
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 3.0.0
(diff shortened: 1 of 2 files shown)


rhysemmas:
Fixed in #310

Issue #245, opened 2019-04-25
Run kiam-server as deployment instead of daemonset


mikhailadvani:
The manifests included in the [deploy](https://github.com/uswitch/kiam/blob/master/deploy/server.yaml) folder as well as the [stable/kiam](https://github.com/helm/charts/blob/master/stable/kiam/templates/server-daemonset.yaml) helm chart both run the kiam-server as a daemonset. While running the kiam-server pods on a dedicated set of nodes, we are left with nodes running only daemonsets. This results in making the recycling of nodes through cordoning/draining difficult.

Proposed Solution:
* Run the kiam-server as a deployment
* Configure it to run with a hard podAntiAffinity
* NodeSelectors/Taints-Tolerations can be used as they are being used with Daemonsets.

thoughts/concerns?

meringu:
What do you find difficult about draining nodes? There is a flag `kubectl drain <node> --ignore-daemonsets`, which will drain the node successfully, but leave daemonsets running. The node can then safely be "recycled", in our case terminated.

iverberk:
We are hitting the exact same issue. The trouble is not with the draining, that works fine, but with the agents not detecting in a small enough window that the kiam servers have been removed. We are upgrading our working in a blue / green fashion, with two ASG. During the drain, the kiam-servers are left untouched because they are running as a daemonset. Of course, on the new nodes they are automatically started. The trouble starts when the old EC2 nodes are terminated. The kiam-servers that were running on those nodes are hard-killed and the kiam-agents seem to have trouble finding the new servers for a very long time.

We've switched the kiam-server from a deamonset to deployment and applied anti-affinity rules to force them on different nodes. There are no observable errors anymore during the upgrade.

However, this still leaves the issue with failing EC2 nodes. EIther we are not configuring the agents properly to detect server failure and fail-over over to the available ones, or the connectivity in the agents is just not handling it well. Either way, would love to get some more insight/help on this.

meringu:
I see. As iam-agent communicateses to kiam server through a Kubernetes service, the kiam-server pods need to be removed.

kiam-agent will also reach a different kiam-server each time, not necessarily on the same node.

Any reason Kiam server needs to run as a daemonset? Only reason I can think of is that the pods should scale proportionally to the cluster, but this could be solved more elegantly with an autoscaler. Changing Kiam server to run as a deployment will not only improve the reliability, but also the cost for larger clusters.

iverberk:
[...]

Issue #217, opened 2019-02-18
kiam server HA. Context deadline exceeded while rotating servers.


max-lobur:
In my test AWS lab I run 3 masters with KIAM servers (`quay.io/uswitch/kiam:v3.0`) and 6 nodes with KIAM agents (same images). I'm getting these 
```
{"addr":"100.112.154.80:56850","level":"error","method":"GET","msg":"error processing request: rpc error: code = DeadlineExceeded desc = context deadline exceeded","path":"/latest/meta-data/iam/security-credentials/","status":500,"time":"2019-02-18T16:48:17Z"}
```
every time I'm rotating masters. The error starts to appear on every one of 6 agents, once I remove at least 1 master. To stop it I need to restart all agents, when they restart they start to work normally. This does not happen if I rotate kiam server itself (delete the pod), only when I rotate the whole master instance.

I'm using this chart https://github.com/helm/charts/tree/master/stable/kiam to run kiam. No special tuning, except for I'm setting `gatewayTimeoutCreation: 500ms` on both agent and server. 

Am I missing smth important in docs, some flag? Please suggest

complex64:
Same issue here, KIAM v.3.2

MilanDasek:
similar issue, kiam v3.3

mehstg:
Just been burned by this myself

mseiwald:
Same issue for us on kiam v3.0

Joseph-Irving:
We run Kiam in relatively large clusters and we regularly terminate our master nodes (we don't allow them to live longer than 7 days) and we've never seen this behaviour.

Kiam should be using a service with `clusterIP: none` this is so that the agents can use grpc loadbalancing, they go and discover the pod ips of each kiam-server and will loadbalance between them, if one goes down it will just try another. So I'm very puzzled as to how you're seeing this. 

I'd be interested to know what networking stack you're using, incase there's some commonality there which could be what's causing this behaviour.

mehstg:
We are using the amazon-k8s-cni plugin for networking. No other service mesh/network policy currently.

denispesikov:
Same here 
`error processing request: error fetching credentials: rpc error: code = DeadlineExceeded desc = context deadline exceeded`
KIAM v3.2
Any workaround?

max-lobur:
Using calico over aws ec2. 
Also using this chart to deploy https://github.com/helm/charts/tree/master/stable/kiam
[...]

PR #296, opened 2019-09-10
Move Server to Deployment instead of DaemonSet

daviddyball:
See discussion in [this issue](https://github.com/uswitch/kiam/issues/245)

Considerations:
* I've set `replicas` to 3, but most will want to adjust this
* I've assumed that antiAffinity is based on pod co-location and not some other criteria
* The `nodeSelector` still ensures it only runs on `master`s, while `spec.affinity` prevents co-location on the same node
* Would it be worth adding some YAML for Horizontal Pod Autoscaling at this stage?

EDIT: Also relates to [issues rolling master nodes causing issues](https://github.com/uswitch/kiam/issues/217)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,18 @@ When your process starts an AWS SDK library will normally use a chain of credent
 ## Deploying to Kubernetes
 Please see the `deploy` directory for example manifests for deploying to Kubernetes.
 
+There are two options for deployment of `kiam-server`, a `Deployment` and a `DaemonSet`. Some users have found the termination handling of the `DaemonSet` pods causes issues when cycling master nodes (see https://github.com/uswitch/kiam/issues/245). Only deploy **one** of the two `server.*.yaml` files, not _both_.
+
+```
+kubectl apply -f deploy/server.deployment.yaml
+kubectl apply -f deploy/service.yaml
+```
+or
+```
+kubectl apply -f deploy/server.daemonset.yaml
+kubectl apply -f deploy/service.yaml
+```
+
 TLS assets must be created to mutually authenticate the agents and server processes; notes are in [docs/TLS.md](docs/TLS.md).
 
 Please also make note of how to configure IAM in your AWS account; notes in [docs/IAM.md](docs/IAM.md).
--- a/README.md
+++ b/README.md
@@ -63,18 +63,6 @@ When your process starts an AWS SDK library will normally use a chain of credent
 ## Deploying to Kubernetes
 Please see the `deploy` directory for example manifests for deploying to Kubernetes.
 
-There are two options for deployment of `kiam-server`, a `Deployment` and a `DaemonSet`. Some users have found the termination handling of the `DaemonSet` pods causes issues when cycling master nodes (see https://github.com/uswitch/kiam/issues/245). Only deploy **one** of the two `server.*.yaml` files, not _both_.
-
-```
-kubectl apply -f deploy/server.deployment.yaml
-kubectl apply -f deploy/service.yaml
-```
-or
-```
-kubectl apply -f deploy/server.daemonset.yaml
-kubectl apply -f deploy/service.yaml
-```
-
 TLS assets must be created to mutually authenticate the agents and server processes; notes are in [docs/TLS.md](docs/TLS.md).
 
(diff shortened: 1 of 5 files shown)


daviddyball:
@rhysemmas @Joseph-Irving : I saw that you had recent approval activity on some closed PRs. Are you active maintainers? Are you able to review this PR?

stefansedich:
Would love to see this one merged, however I wonder @daviddyball if it should provide an option to either run as a DaemonSet or a Deployment depending on what a user prefers.

daviddyball:
Very true @stefansedich. It's currently not a helm template, so adding some sort of "switching" logic is not really possible. Perhaps just offering two different YAML files is enough? My only concern would be people blindly doing `kubectl apply -f deploy/*.yaml` :stuck_out_tongue:

stefansedich:
Ah sorry @daviddyball it is early and I failed to realize this was not the helm chart!

denniswebb:
If this has gotten stale, I'm open to doing the helm chart work because as it stands, I'm about to have to start hosting the chart myself to do the daemonset vs deployment work. Please ping me if I'm open to do the helm chart work myself on a separate PR.

daviddyball:
This isn't exactly stale, it's just not gotten reviewed by any maintainers :confused: We've been running `kiam-server` as a `Deployment` since September and it's solved most of our issues with masters being rolled.

jaygorrell:
I'm curious if the maintainers have any opposition to this approach. The chart PR linked above (https://github.com/uswitch/kiam/pull/332) adds an _option_ for moving to Deployment without changing the default... so hopefully that can get accepted more easily.

If the default is changing, the chart should change to match but also there are things like the README that should be updated as part of this PR. We just need input from a maintainer though.

daviddyball:
This one doesn't force a default, it just offers a different YAML file if you _choose_ to run it as a `Deployment` :man_shrugging:

jaygorrell:
It's not a different YAML file though - it's changing the `DaemonSet` file to be `Deployment`. 

I have no authority here, just saying it may be confusing to have the README talk about using a `DaemonSet` while the example in `deploy/server.yaml` is a `Deployment`. I like the change, but updating the README would make sense.

daviddyball:
I had it in my head that I'd already refactored this to make it optional which deployment method you choose... turns out I hadn't :roll_eyes: 

Moved to 2x YAML file options under `deploy/`, `server.daemonset.yaml` and `server.deployment.yaml` and updated `README.md` to include instructions and a note about only deploying _one_ of the two options available.

jaygorrell:
Looks good to me. I'd approve if I could. :)

README.md
```diff
@@ -63,6 +63,18 @@ When your process starts an AWS SDK library will normally use a chain of credent
...
+
+```
+kubectl apply -f deploy/server.deployment.yaml
+kubectl apply -f deploy/service.yaml
+```
+or
+```
+kubectl apply -f deploy/server.daemonset.yaml
+kubectl apply -f deploy/service.yaml
+```
```
Joseph-Irving: I would probably avoid directly telling people to kubectl apply these manifests, they were intended as examples. They're not for production use, like they use latest as the tag and potentially need you to change some things such as where you mount certs in from. So I think I'd just call out that there are two ways to deploy and you can find them in the deploy folder if you want examples

jaygorrell: Not that I disagree, but this does contextually fit with what was already here.

daviddyball: Removed

deploy/server.deployment.yaml
```diff
@@ -0,0 +1,99 @@
...
+kind: Deployment
+metadata:
+  namespace: kube-system
+  name: kiam-server
+  labels:
+    app: kiam
+    role: server
+spec:
+  replicas: 3
+  selector:
```
dbyron0: I think this is missing:
```
  strategy:
    rollingUpdate:
      maxSurge: 0
      maxUnavailable: 1
    type: RollingUpdate
```
as the helm chart has [here](https://github.com/uswitch/kiam/blob/master/helm/kiam/templates/server-deployment.yaml#L14)

See https://github.com/uswitch/kiam/pull/332/files#r351312455 for background.

PR #306, opened 2019-10-13
Update aws-sdk-go

dcherman:
In order to use the new IAM Roles for Service accounts feature from AWS,
kiam needs to use a newer version of the AWS SDK.

Using a service account allows you to avoid the need to give the entire node
AssumeRole permissions via a node role.  This eliminates the potential for pods to be scheduled onto the kiam server nodes and get access to the kiam role.  I've also noticed that if you turn off role base arn autodetect and explicitly provide it, this allows you to run the server/agents on the same node successfully.

While in an ideal world we would not need to combine the use of kiam with the official AWS solution, there are multiple projects that still need to update their dependencies in order to use that new functionality.


rhysemmas:
Thanks for contributing @dcherman - I am closing this PR as we are managing dependencies in the project with Go mod rather than Dep now and the SDK updates have been addressed in #330

PR #307, opened 2019-10-14
Update helm daemonsets apiversion for 1.16 support

velothump:


velothump:
Thanks @rhysemmas . I didn't bump the version as the Changelog looked autogenerated. Let me know if any more changes are needed.

Issue #308, opened 2019-10-14
Taint node if agent is unable to connect to server 


zytek:
Kind of bug report / kind of feature request ;-)

We have an issue with kiam agent failing to connect to kiam server randomly. Logs are not very verbose:
`{"level":"fatal","msg":"error creating server gateway: error dialing grpc server: context deadline exceeded","time":"2019-10-10T11:05:36Z"}` 
Kiam-agent then exits and is restarted by Kubernetes, still fails and goes into `CrashLoopBackOff`. This makes a node unusable by apps requiring IAM access.

The issues is hard to debug as logging to some other pod on the same node and trying to `curl` to the kiam-server works fine. Other nodes in the cluster with kiam-agent work fine also. 

My take is that it's either a kubernetes networking / sandboxing issue or it might be related to #299 

Apart from actually fixing the problem I was thinking about implementing a watchdog mechanism and I am seeking your guidance.  My initial idea was to deploy node-problem-detector and write a custom check for kiam-agent. Hacky, but should work.

My another idea would be to add ability to taint node by kiam-agent itself when it experiences problems with communication to kiam-server. Users would then be able to tolerate such taints in workflows than can work without AWS credentials. Other pods that require working `kiam` could be then evicted and run on other nodes. 

What do you think about such feature ? Should you think that it could be a part of kiam-agent I would be happy to work on a PR with my team, but first would like to get your guidance. This would require changing current behaviour of exiting in case of connection issue to a retry with backoff mechanism and some logic to taint a node and maybe use Events / Node Condition API for additional verbosity.

zytek:
FYI kiam problems were probably related due to poor kiam chart defaults, mainy:
- lack of resource definition leading to CPU starvation on busy node
- low gateway-timeout-creation

PR #310, opened 2019-10-17
helm chart: update default gatewayTimeoutCreation

zytek:
Bump default to 500ms, as 50ms leads to easy connection issues on busy nodes, especially when resource definitions are left empty so kiam is easily starved.

I would also propose to bump it here https://github.com/uswitch/kiam/blob/master/cmd/kiam/opts.go#L117

Related: #308

### DIFF
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 4.1.0
+version: 4.2.0
 appVersion: 3.4
 description: Integrate AWS IAM with Kubernetes
 keywords:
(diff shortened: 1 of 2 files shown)


zytek:
Uhm, duplicate of #294

helm/kiam/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 4.0.0
+version: 4.1.0
```
rhysemmas: Sorry - #316 merged previously has bumped the version to `4.1.0` already, can we bump to `4.2.0` please

zytek:
@rhysemmas  rebased, push-forced

PR #311, opened 2019-10-17
Increase default gateway-timeout-creation

zytek:
On busy nodes kiam gets easily starved out of CPU and connections to kiam-servers fail with `context deadline exceeded`.

Related: #308 #310

### DIFF
--- a/cmd/kiam/opts.go
+++ b/cmd/kiam/opts.go
@@ -114,7 +114,7 @@ type clientOptions struct {
 func (o *clientOptions) bind(parser parser) {
 	parser.Flag("server-address", "gRPC address to Kiam server service").Default("localhost:9610").StringVar(&o.serverAddress)
 	parser.Flag("server-address-refresh", "Interval to refresh server service endpoints ( deprecated )").Default("0s").DurationVar(&o.serverAddressRefresh)
-	parser.Flag("gateway-timeout-creation", "Timeout to create the kiam gateway ").Default("50ms").DurationVar(&o.timeoutKiamGateway)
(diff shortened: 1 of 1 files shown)


rhysemmas:
I think having a low default value for the gateway timeout flag is fine - we ideally want to keep this value as low as possible and as we overwrite the default using the manifests or the helm values when deploying to Kube anyway, it feels a bit unnecessary to adjust the timeout here. 

Also, I don't think we should adjust the timeout based on the Kiam agent being starved of CPU, as this feels like it could turn into a never-ending cycle of increasing timeouts because of resource issues on a node. CPU starvation can easily be resolved by setting sufficient resource requests in the Kiam daemonsets.

zytek:
I tend disagree - as in my opinion defaults should be as reliable as possible and we should leave **optimisation** (lower timeouts) to the user, not the other way around. You can see from mine and others issues that CPU starvation on shared infastructure, on such low defaults, is quite common.

But changing defaults in helm chart should be sufficient.

rhysemmas:
Fair point, my main concern was that we are increasing a timeout because of issues with CPU starvation on nodes and my thinking was that the solution to this is to address the lack of resource requests for Kiam pods, rather than increasing a timeout. 

I suppose the thought process for my original point of keeping the timeout low in the flag stemmed from Kiam not running in Kube and therefore 50ms being a reasonable gateway timeout default for local testing. As we would overwrite the default when deploying to Kube, it seemed unnecessary to change the default in the flag - but saying that, it would make sense for the default to reflect where Kiam will spend most of its time running.

I'm happy to add 1s as the default for the flag if you'd like to update the PR, to keep the value consistent across the repo. It will be included in a later release of Kiam.

zytek:
@rhysemmas updated

PR #319, opened 2019-11-04
Helm: adding optional deep health check to the agent's liveness probe

stanvit:
Implementing deep health check feature (#268) in the helm chart .

Agents are restarted when communication with masters fails, which may happen on certificate changes.

### DIFF
--- a/helm/kiam/CHANGELOG.md
+++ b/helm/kiam/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Helm Chart Changelog
 
+## 5.1.0
+8 November 2019
+
+Notable changes:
+* [#319](https://github.com/uswitch/kiam/pull/319) Optional [deep liveness check](https://github.com/uswitch/kiam/pull/268) has been added to the helm chart.
+
+Many thanks to the following contributor for this release:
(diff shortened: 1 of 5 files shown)


stanvit:
@rhysemmas , I've updated `Chart.yaml` as well as the cart's `CHANGELOG.md`, as it seems that you prefer contributors to do that themselves.

helm/kiam/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Helm Chart Changelog
 
+## 5.1.0
+8 November 2019
+
+Notable changes:
+* [#319](https://github.com/uswitch/kiam/pull/319) Otional [deep liveness check](https://github.com/uswitch/kiam/pull/268) has been added to the helm chart.
```
rhysemmas: Missing 'p' from Optional

rhysemmas:
Thanks @stanvit, please see my comment above

helm/kiam/CHANGELOG.md
stanvit: Fixed

PR #320, opened 2019-11-05
Set default SSL host path for helm template. Add docs on SSL host pat…

MVJosh:
- Improve documentation on key parameters required for helm install to work.
- Set default SSL host path for helm template to match example in `deploy/server.yaml`


helm/kiam/README.md
```diff
@@ -27,6 +27,13 @@ $ helm install uswitch/kiam --name my-release
 
 The command deploys kiam on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation.
 
+## Key Configuration
+The default helm configuration will probably not work out-of-the-box. You will most likely need to adjust the following:
+
+* Kiam will _not_ work without an appropriate iptables rule. As there are security & operational implications with making kiam responsible for inserting & removing the rule (see [#202](https://github.com/uswitch/kiam/issues/202) & [#253](https://github.com/uswitch/kiam/pull/253)), the `agent.host.iptables` parameter is set to `false` by default. Either configure the iptables rule separately from Kiam, or use `--set agent.host.iptables=true`.
+* Kiam requires SSL certs from the host system to be mounted into the kiam-server pod in order to be able to contact the AWS meta-data API. If the SSL cert directory on the host(s) you intend to run the kiam-server on does not match the default (`/usr/share/ca-certificates`), you will need to set the `server.sslCertHostPath` variable.
+
```
Nuru: Suggested extra material to explain `iptables` rule:

```suggestion
* Kiam requires SSL certs from the host system to be mounted into the kiam-server pod in order to be able to contact the AWS meta-data API. If the SSL cert directory on the host(s) you intend to run the kiam-server on does not match the default (`/usr/share/ca-certificates`), you will need to set the `server.sslCertHostPath` variable.
* Kiam will _not_ work without an appropriate iptables rule. As there are security & operational implications with making kiam responsible for inserting & removing the rule (see [#202](https://github.com/uswitch/kiam/issues/202) & [#253](https://github.com/uswitch/kiam/pull/253)), the `agent.host.iptables` parameter is set to `false` by default. Either configure the iptables rule separately from Kiam, or use `--set agent.host.iptables=true`.

### Adding `iptables` rule separately
The most secure way to configure `kiam` is to install its `iptables` rule before starting Docker on the host, and leave it in place. This way when `kiam` is not able to serve credentials, any clients attempting to refresh credentials will get an error, and they should continue to use their cached credentials while periodically retrying to refresh them. Likewise clients attempting to get credentials for the first time will get nothing, rather than get the credentials associated with the host they are running on.

You can [read the code](https://github.com/uswitch/kiam/blob/master/cmd/kiam/iptables.go), but the definitive way to see the `iptables` rule `kiam` installs is to `--set agent.host.iptables=true`, deploy the agent, find the pod name for an agent pod and exec into it:
`
kubectl exec -it kiam-agent-qhtgr -- iptables -t nat -S PREROUTING | grep 169.254.169.254/32
`
This will print out the arguments to follow `iptables -t nat` needed to install the rule. However, the arguments will include the IP address of the machine, so you cannot just copy and paste them verbatim. You can, however, use it to double-check you have the correct rule installed. 

This command works on `debian` hosts on AWS when using `calico` networking:
`
/sbin/iptables -t nat -A PREROUTING -d 169.254.169.254/32 \
        -i cali+ -p tcp -m tcp --dport 80 -j DNAT \
        --to-destination $(curl -s http://169.254.169.254/latest/meta-data/local-ipv4):8181
`     
Replace `cali+` with the CNI interface name you are passing to `kiam-agent` as `--host-interface` and if you are using exclamation point to invert it, e.g. `!eth0` be sure to escape the exclamation point so it is not interpreted by the shell. 

See [this example](https://github.com/cloudposse/reference-architectures/blob/5cad9b5d97dd3e92257ca2ef37ca6b8ff96940d0/templates/kops/kops-private-topology.yaml.gotmpl#L137-L156) for how to install the `iptables` rule using a `systemd` unit installed as a `kops` install hook.

It is safe to have the rule installed twice, so you can test your installation by installing the rule once via the `kiam-agent` and once some other way, in which case you should see the same rule installed twice when you print out the rules. If the 2 rules are identical, then you have the install correct and can revert `agent.host.iptables` to false. 

```

helm/kiam/README.md
```diff
@@ -27,6 +27,35 @@ $ helm install uswitch/kiam --name my-release
...
+This command works on `debian` hosts on AWS when using `calico` networking:
+```bash
+/sbin/iptables -t nat -A PREROUTING -d 169.254.169.254/32 \
+        -i cali+ -p tcp -m tcp --dport 80 -j DNAT \
+        --to-destination $(curl -s http://169.254.169.254/latest/meta-data/local-ipv4):8181
+```
+Replace `cali+` with the CNI interface name you are passing to `kiam-agent` as `--host-interface` and if you are using exclamation point to invert it, e.g. `!eth0` be sure to escape the exclamation point so it is not interpreted by the shell. 
+
+See [this example](https://github.com/cloudposse/reference-architectures/blob/5cad9b5d97dd3e92257ca2ef37ca6b8ff96940d0/templates/kops/kops-private-topology.yaml.gotmpl#L137-L156) for how to install the `iptables` rule using a `systemd` unit installed as a `kops` install hook.
```
rhysemmas: Thanks @MVJosh and @Nuru for contributing - my only thought was that it would be nice if we could include the systemd service example for adding the kiam iptables rule manually in this repo. The linked file contains a lot of other stuff which looks a bit messy surrounding the small example that we need and it means we will have control to change the example if needed. Could include something like `taken from cloudposse/reference-architectures/...` so people who are interested can look up the complete kops install hook.

helm/kiam/README.md
```diff
@@ -27,6 +27,35 @@ $ helm install uswitch/kiam --name my-release
 
 The command deploys kiam on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation.
 
+## Key Configuration
+The default helm configuration will probably not work out-of-the-box. You will most likely need to adjust the following:
+
+* Kiam requires SSL certs from the host system to be mounted into the kiam-server pod in order to be able to contact the AWS meta-data API. If the SSL cert directory on the host(s) you intend to run the kiam-server on does not match the default (`/usr/share/ca-certificates`), you will need to set the `server.sslCertHostPath` variable.
```
Nuru: Clarify that this is adding globally trusted root certificates, not kiam-specific certificates.
```suggestion
* Kiam requires trusted SSL root certificates from the host system to be mounted into the kiam-server pod in order to be able to contact the AWS meta-data API. If the SSL cert directory on the host(s) you intend to run the kiam-server on does not match the default (`/usr/share/ca-certificates`), you will need to set the `server.sslCertHostPath` variable.
```

helm/kiam/README.md
Nuru: > ...my only thought was that it would be nice if we could include the systemd service example for adding the kiam iptables rule manually in this repo.

@rhysemmas My thought on that is (1) this is a README and should be kept short. I would welcome a separate document that goes into detail about these kind of configuration options and how `kiam` works, but I prefer README to be short so that people do not get overwhelmed by it, and (2) the key piece of documentation is the `iptables` command, and people should be able to figure out what they need on their cluster from there. We do not really need to include a `systemd` unit or a `kops` install hook, and we do not have those thoroughly vetted or officially supported by the `kiam` team, so we should not directly include them in the documentation. I am particularly concerned that with the large number of possible combinations of host OS and CNI implementations, the `Before` and `After` clauses may not be correct or sufficient for some cases. If it turns out there is a problem with the linked example, well, it was an external example, not an official `kiam` recommendation. 

If/when it comes time that the linked example needs to be changed, at least then we will have some amount of vetting and official support in place (to drive the change) and at that point the link can be replaced with vetted code. At that point it is a tiny bit of code and no need to maintain a "taken from" reference.

Joseph-Irving: I'm not sure linking to that example adds much to this readme, I think it would be fine without it. As you said there a lot of different OSes, CNIs etc, so I'm not sure having a linking off to a systemd unit for a very specific use-case, contained in an unrelated repo, that we have no control over really adds much to this document.

helm/kiam/README.md
MVJosh: Good spot, thanks!

helm/kiam/README.md
Nuru: @Joseph-Irving Fine with me if you want to leave the link out. It’s just that without it the README gives no guidance for how to actually get the `iptables` command executed on the node instances when they are created. I’m trying to give some direction, while acknowledging the fact I cannot give a comprehensive and definitive answer.

MVJosh: What do we want to do about this? I'm happy to include the example Nuru linked to directly in the readme here (with a reference to the original) if that's best

rhysemmas: Sorry for the delay getting back to this, my opinion is that I think we can remove the example for setting up the iptables rule with a systemd unit. I think that the type of user that does not already know how to add an iptables rule to their host is unlikely to find the example very useful without a complete and supported explanation for doing this on their own setup, the likelihood is that they will just let Kiam add the rule for them. 

I think this section of the README highlights an important point - adding the iptables rule manually - and those who feel so inclined can look into the best way to do it for themselves, until we can provide a complete example/guide on doing this on common setups.

MVJosh: Agreed - I think the additional detail around how to determine what the rule should be is enough for now.

rhysemmas:
Thanks for this @Nuru and @MVJosh - please can you resolve the conflicts (looks like the chart version just needs bumping) and we will merge this in

PR #322, opened 2019-11-06
helm: add support for iptables-remove flag

zytek:
Suport iptables-remove flag and point users to potential use-case. 

Signed-off-by: Jakub Paweł Głazik <zytek@nuxi.pl>

### DIFF
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 4.2.0
+version: 4.2.1
 appVersion: 3.4
 description: Integrate AWS IAM with Kubernetes
 keywords:
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 4.2.1
+version: 5.0.0
 appVersion: 3.4
 description: Integrate AWS IAM with Kubernetes
 keywords:
--- a/helm/kiam/templates/agent-daemonset.yaml
+++ b/helm/kiam/templates/agent-daemonset.yaml
@@ -81,6 +81,9 @@ spec:
             {{- if .Values.agent.host.iptables }}
             - --iptables
             {{- end }}
+            {{- if not .Values.agent.host.iptablesRemoveOnShutdown }}
+            - --no-iptables-remove
+            {{- end }}
             - --host-interface={{ .Values.agent.host.interface }}
             {{- if .Values.agent.log.jsonOutput }}
             - --json-log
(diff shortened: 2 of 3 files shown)


rhysemmas:
Thanks for this @zytek, could we please bump the minor chart version (rather than patch), as a backwards-compatible functionality change is being added

helm/kiam/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 4.2.0
+version: 4.2.1
```
rhysemmas: Please bump to `4.3.0`

Nuru:
I am strongly in favor of adding this option to the chart, but still I have changes to request:

1. Keep the sense of the helm option the same as the command-line option by changing the name from `iptablesKeepOnShutdown` to `iptablesRemoveOnShutdown` and adding the `--no-iptables-remove` flag when `iptablesRemoveOnShutdown` is false
2. Default `iptablesRemoveOnShutdown` to false

As discussed in #202, `--no-iptables-remove` is the safer option from a security standpoint, and is required to enable RollingUpdate deployment strategy to work. The only downside to having `--no-iptables-remove` comes when and if you fully remove `kiam` from your cluster without intending to re-install it, which is a very rare occurrence, _and_ you had turned on `agent.host.iptables`, which is not the default. If desired, this could be explained in a separate section of the README about uninstalling `kiam`, which remains slightly complicated by the `iptables` rules no matter what settings you set. 

If you make these changes, then because it changes the default of `iptables-remove` I think it should be a minor version upgrade, not a patch version upgrade.

zytek:
I've renamed the flag and set default value to `false` - I agree that from usability (RollingUpdate) and security standpoint this is the correct behaviour. By default iptables rule should be kept after kiam exits, if the user wishes to add it in the first place. 

Chart version bumped to 5.0.0 as this changes default behaviour.

Side note: I believe chart versions should not be touched by PRs and altered only by person making merges and updating changelog - as older PR's are forced to have conflicts and it's and never ending race between old and new PRs.

rhysemmas:
Thanks for this @zytek and @Nuru

In regards to the chart versions, we have a CI step that packages and publishes the chart to our charts repo once the code is merged to master, so it is necessary to have the chart version incremented in the PR so that when it is merged we don't overwrite whatever the last chart was. 

It is unfortunate that the chart version needs to be bumped again if another PR is merged prior, we will try to reduce the need for this by reviewing PRs in the order that they come in but that won't always work as PRs which require less changes or discussion will get merged more quickly.

This is the way it currently works in the official Helm charts repo, I am open to hearing suggestions on how we can improve this workflow, though, if anyone has some ideas.

As an aside: I've been updating the changelog after PRs get merged as it usually not included in PRs that contributors raise, which I appreciate is due to our lack of contributing guidelines and I will address this soon. Until then, if you'd like to update the changelog in future contributions that would be a considerable help and please do make sure to credit yourselves and fellow contributors for your great work!

Issue #326, opened 2019-11-11
Unable to limit by IAM path on namespace with RoleBaseARN


leosunmo:
All of my Kubernetes workloads have roles created under the IAM path `workloads`. I am running in to an issue where I can't use regex to require `/workloads/.*` prefix on the roles to be assumed.
The problem I think I am running in to is the `role-base-arn` isn't appended until _after_ the policy check.

Here's my current setup.

The role I am trying to assume in my Pod is `arn:aws:iam::1234567890:role/workloads/my-amazing-role`

The Kiam server has a role base ARN set
```
--role-base-arn=arn:aws:iam::1234567890:role/workloads/
```

My namespace has the following regex:
```
annotations:
  iam.amazonaws.com/permitted: .*workloads/.*
```

and finally my test pod has this config:
```
annotations:
  iam.amazonaws.com/role: my-amazing-role
```
This looks like it should work from looking at how the `role-base-arn` and `iam.amazonaws.com/permitted` config works.

The server gives me the following error:
```
  Warning  KiamRoleForbidden  8s    kiam.serve    failed assuming role "my-amazing-role`": namespace policy expression '.*workloads/.*' forbids role 'my-amazing-role'`
```

It looks like the full ARN is not resolved at the time of policy evaluation. I think I found the fix in the following old issue, https://github.com/uswitch/kiam/issues/75. @roffe's fix at the end of that issue _should_ fix my issue.
The fix is at https://github.com/roffe/kiam/blob/5f04cecbef761bc0a6a3ba5141497c1ff9585c54/pkg/server/policy.go#L149 but was never merged in to master, https://github.com/uswitch/kiam/blob/master/pkg/server/policy.go#L149.

As far as I understand this means it evaluates the "role" that the Pod annotation provides against the regex in the namespace, not the expanded/resolved role ARN.

leosunmo:
Just made a quick hack and got it to work.

Fork: https://github.com/uswitch/kiam/compare/master...leosunmo:arn-resolve-before-policycheck
Docker image that is working in my environment right now: https://hub.docker.com/repository/docker/leosunmo/kiam

I have not fixed the tests yet as I am not as familiar with how these tests are written. I'll try to take a look as soon as possible.

leosunmo:
PR brought in line with Master and tests have been written.

leosunmo:
Any chance of having this looked at? I'd really like to move away from running a fork in my production environment!

pingles:
[...]

PR #327, opened 2019-11-14
Resolve base ARN before checking policy

leosunmo:
This PR ensures that we resolve the role ARN before we check if it's allowed in the Namespace white-list/regex.

Resolves: #326 

This will allow you to use subpaths in the base ARN and match against roles using wildcards without having to include the IAM subpath. More details in original issue.


leosunmo:
This has been brought in line with Master now. Still working well in my production environment.

leosunmo:
This has once again been brought inline with master. Still running this in my production environment and it's working well.

leosunmo:
@pingles This might be one for v4 as well.

pingles:
Sounds good, thanks! Will take a look.

pingles:
Having a quick look, I wonder whether it makes sense to simplify both policies and for them to only operate on the full resolved role, rather than passing the resolver in as an argument?

It's been a while since I've looked at the rest of the code so not sure of the implications but feels like responsibilities aren't quite right at the moment.

Appreciate it's probably been a while for you too @leosunmo but what do you think?

leosunmo:
I totally agree. Passing the resolver was always the way way out :P I just didn't want to make major changes or spend too much time on it as I wasn't sure if it would be merged. 

With a new release it makes more sense to make some more fundamental changes. I would have to spend some time understanding it first though.

pingles:
Cool. If you don’t get a chance I’ll try and pick it up, thanks for sharing
here though.

leosunmo:
@pingles Has this been incorporated in to https://github.com/uswitch/kiam/pull/435 ?

pingles:
@leosunmo yep! thanks for contributing :-) I made a few changes to fit in with some of the other refactorings I was doing at the time but all good.

pingles:
I'll close this given #435 is merged and had your commits. Thanks again

Issue #328, opened 2019-11-14
Namespace whitelist annotation is too permissive


leosunmo:
I came across this while writing tests for https://github.com/uswitch/kiam/pull/327. 

The default regex at https://github.com/uswitch/kiam/blob/master/pkg/server/policy.go#L164 is far too permissive. 

Unless I am missing something, I am able to make it match with most things as long as that character appears once in the role string. For example: 
https://goplay.space/#w4k3fBwrXVJ

Playing around here https://regex101.com/r/a2eGqY/1/  
you can see that we'd have to detect if the string has any regex characters and if it doesn't, add something like `^i$` to make it not match a single character in the role.

Not sure if there's a nicer way around the regex issue or if it requires some string analysing to figure out how to construct the regex query.

PR #329, opened 2019-11-14
Strict namespace annotation regex

leosunmo:
This (rather crudely) adds beginning-of-string and end-of-string assertion for the namespace annotation regex. 

Resolves: #328 

This should stop really short or vague regex from passing.

I was afraid regex with pre-existing assertions might cause issues (https://github.com/leosunmo/kiam/blob/978092d031db57b4d7d884d1956822e2075e65d2/pkg/server/policy_test.go#L124), but it doesn't look like it. I had a good play around [here](https://regex101.com/r/OIH8Gd/1) and it seems fine to double-up on these assertions.

We could always check if the expression string starts and/or ends we these assertions and trim it.


pingles:
Thanks for adding this @leosunmo, we'll try and take a look asap.

Joseph-Irving:
this seems like a very sensible change, however there is the possibility that there are people currently relying on this "feature" so making it stricter could be a breaking change for them. Maybe we should add this behind a flag and look at making it default in the next major version?

pingles:
Agreed. I'd potentially control with a flag so it's also possible for people to disable if they prefer, but agreed, better to default to current behaviour for now and potentially change on a major release.

leosunmo:
Pulling in release v3.5. Didn't seem to change anything, will re-run tests.

leosunmo:
Tests still pass. At some point I could look at adding a flag, but I don't have a lot of free time to look at this.

leosunmo:
This has been brought in line with master branch.

Added the flag `--strict-namespace-regexp` that enabled this behaviour. It defaults to disabled which should shield anyone from breaking behaviour except for people explicitly encountering this issue, such as myself.

Let me know if the flag makes sense, the code looks good and the tests are sane. Still running this change (and https://github.com/uswitch/kiam/pull/327) in production right now, and it's working well.

pingles:
I've been teeing up other changes for v4 that makes some breaking flag and deprecation changes. While reviewing issues and PRs I came back to this.

@Joseph-Irving @leosunmo how do you feel about incorporating now? I feel like the default behaviour should now be the stricter form with the ability for people to revert to v3.x behaviour if necessary?

leosunmo:
I think making this a breaking change with a revert flag is a great idea. I will try to find time to look at this again asap.

leosunmo:
Okay, I've made the strict namespace regex behaviour default. The flag is a bit cumbersome but at least it's explicit and clear about its behaviour.

Inverted the tests to hopefully catch the new expected default behaviour.

pingles:
Thanks again @leosunmo. I'm still keen to pull this into v4 (I'll mark it for the milestone) and will try and help with the merge (there's been a few other changes to help improve things and pull in ExternalID / Session Naming support).

leosunmo:
@pingles Okay went through and fixed merge conflicts.
It all ended up being one giant merge commit unfortunately, but I updated test behaviour (making "strict namespace regexp" the default behaviour), added some comments and renamed tests for clarity.
Also simplified the naming around disabling strict regexp for namespaces.

Let me know if there's something you'd like to see changed.

pkg/server/server_builder.go
```diff
@@ -209,7 +209,7 @@ func (b *KiamServerBuilder) Build() (*KiamServer, error) {
 		credentialsProvider: credentialsCache,
 		assumePolicy: Policies(
 			NewRequestingAnnotatedRolePolicy(b.podCache, arnResolver),
-			NewNamespacePermittedRoleNamePolicy(b.namespaceCache, arnResolver),
+			NewNamespacePermittedRoleNamePolicy(!b.config.DisableStrictNamespaceRegexp, b.namespaceCache, arnResolver),
```
leosunmo, line 212: Generally don't like subtle things like this where you end up with a double-negative...
I felt like the code should deal with a `strict` bool that was by default true, but the flag should be very verbose to make sure you understand you are about to reduce safety.

pingles (approved):

PR #331, opened 2019-11-22
enable gRPC keepalive to detect dead TCP connections (#217)

mechpen:
When a TCP connection dies, i.e. no more packet from peer, gRPC
does not close the connection by default.  We need to enable gRPC
keepalive to detect and close dead TCP connections.

The error can be easily reproduced with iptables command on the agent host:
```
iptables -I INPUT -s <server-ip> -p tcp --sport 443 -j DROP
```


mechpen:
also opened an issue with grpc-go: https://github.com/grpc/grpc-go/issues/3206

mechpen:
need approval from my company

PR #332, opened 2019-11-22
[helm] adds option to run server as deployment

denniswebb:
This relates to #296 and #245 

This provides the ability to run the server as a deployment in the helm chart.

### DIFF
--- a/helm/kiam/templates/server-deployment.yaml
+++ b/helm/kiam/templates/server-deployment.yaml
@@ -1,6 +1,6 @@
-{{- if and .Values.server.enabled ( eq .Values.server.deployment.enabled false) -}}
+{{- if and .Values.server.enabled .Values.server.deployment.enabled -}}
 apiVersion: apps/v1
-kind: DaemonSet
+kind: Deployment
 metadata:
   labels:
     app: {{ template "kiam.name" . }}
@@ -10,6 +10,7 @@ metadata:
     release: {{ .Release.Name }}
   name: {{ template "kiam.fullname" . }}-server
 spec:
+  replicas: {{ .Values.server.deployment.replicas }}
   selector:
     matchLabels:
       app: {{ template "kiam.name" . }}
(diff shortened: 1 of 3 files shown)


jaygorrell:
Help address https://github.com/uswitch/kiam/issues/217 as well

helm/kiam/templates/server-deployment.yaml
```diff
@@ -0,0 +1,168 @@
...
+{{ toYaml .Values.server.tolerations | indent 8 }}
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - labelSelector:
+                matchExpressions:
+                  - key: "app"
+                    operator: In
+                    values:
+                      - {{ template "kiam.name" . }}
```
rhysemmas: It looks like this label exists so that server pods are not colocated on nodes running agents. I don't think we should add this in as a default, as it is possible to colocate agent and server pods on nodes without issue and this default would take that option away from users. The current accepted practice to keep agent and server pods separated is to have kiam server pods scheduled onto dedicated nodes or masters through the use of nodeSelector (and tolerations). 

Adding antiaffinity to the server pods with the labelSelector below seems reasonable as this keeps the daemonset behaviour of scheduling only one server pod on a node. I think we should continue to follow the recommended way of separating agent and server pods, though, using nodeSelector.

As an aside - I'm not sure that this is the correct syntax to add 2 labels for antiaffinity, as in my testing I still see server pods colocated on nodes with agent pods (both pods with the "app: kiam" label).

rhysemmas:
Many thanks for this @denniswebb - please see my comment and let me know your thoughts

helm/kiam/templates/server-deployment.yaml
denniswebb: Looking [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) shows that these are ANDed and is only preventing server pods from being co-located. I do know this works because when I had rollingUpdates not allow any unavailable pods, it got hung because with only 3 nodes it was unable to start replacement pods.
This is why you are able to see server and agent pods running together in your configuration. With ours, we use nodeSelectors and taints to prevent colocating of server and agent pods. We run nodes purely for the purpose of hosting server pods.
If I'm missing the point of your comment let me know, but I think I read and understand your concerns correctly.

rhysemmas: Great - thanks for the explanation and the link, I was struggling to find an example of two label matches in affinity and with the behaviour I was seeing assumed they were being ORed. Please bump the chart version (current version is now 5.2.0) and we will merge this in.

denniswebb: Version bumped. Thanks

PR #333, opened 2019-11-25
Issue #217 - Add GRPC keepalive to server/client

mehstg:
This PR adds a keepalive between servers/clients ensuring connections are closed correctly in the case of ungraceful KIAM server termination.

For reference:
Kiam issue - https://github.com/uswitch/kiam/issues/217
Go-GRPC issue: https://github.com/grpc/grpc-go/issues/3206

Putting this forward for review. I am currently running this patch in our main development cluster with no ill effects. I am not a golang dev by any stretch of the imagination so am not sure of the knock on effects of enabling this.

### DIFF
--- a/pkg/server/gateway.go
+++ b/pkg/server/gateway.go
@@ -52,6 +52,12 @@ const (
 	RetryInterval = 10 * time.Millisecond
 )
 
+var kacp = keepalive.ClientParameters{
+	Time:                10 * time.Second, // send pings every 10 seconds if there is no activity
+	Timeout:             time.Second,      // wait 1 second for ping ack before considering the connection dead
+	PermitWithoutStream: true,             // send pings even without active streams
+}
+
 // NewGateway constructs a gRPC client to talk to the server
 func NewGateway(ctx context.Context, address string, caFile, certificateFile, keyFile string) (*KiamGateway, error) {
 	callOpts := []retry.CallOption{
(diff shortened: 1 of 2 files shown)


Joseph-Irving:
Hi thanks for this contribution, it would be nice if these grpc options could be configured by flags, having sane defaults is always good but if people need to tweak or turn off this functionality it would be good to have the option.

pkg/server/gateway.go
```diff
@@ -52,6 +53,12 @@ const (
 	RetryInterval = 10 * time.Millisecond
 )
 
+var kacp = keepalive.ClientParameters{
```
Joseph-Irving: nit: I generally prefer to have variable names that are long and obvious instead of short and obscure, so `keepAliveClientParams` would be preferable over `kacp`.

mehstg: Have updated the vars

mehstg:
> Hi thanks for this contribution, it would be nice if these grpc options could be configured by flags, having sane defaults is always good but if people need to tweak or turn off this functionality it would be good to have the option.

I'm afraid this is probably a little beyond my limited golang skills. Happy for someone else to pick it up though.

dbyron0:
https://github.com/uswitch/kiam/pull/337 is enough to replace this?

Joseph-Irving:
thanks for trying this @mehstg!
this PR has implemented the same thing #337

PR #336, opened 2019-11-30
Helm service account annotations and fixes for deployments

tyrken:
* Annotations to allow agent/service service-account annotations
* Fixes for deployment object spec, removing erroneous updateStrategy
* Add (back?) the mount of the host's SSL certs for deployments

I'm using these to work with EKS IAM roles for services accounts... do they look ok?

### DIFF
--- a/helm/kiam/CHANGELOG.md
+++ b/helm/kiam/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Helm Chart Changelog
 
+## 5.4.0 (unreleased)
+
+Notable changes:
+* [#336](https://github.com/uswitch/kiam/pull/336) Add optional service-account annotations for server and agent.  Fix deployment to work with helm v3 and custom SSL host path.
+
 ## 5.3.0
 27 November 2019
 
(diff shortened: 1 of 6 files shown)


sbkg0002:
+1

The current version doesn't work with Helm 3:
```
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(Deployment.spec): unknown field "updateStrategy" in io.k8s.api.apps.v1.DeploymentSpec
```

rhysemmas:
Thanks for this @tyrken! LGTM, but could you please bump the chart version in `Chart.yaml` so a new chart is packaged when we merge. I believe this would be a minor release, as a backwards-compatible functionality change is being made (alongside the bug fixes)

To expand on the removal of `updateStrategy` from the deployment template for anyone that comes across this PR: the `updateStrategy` field is only valid for daemonsets (which is the default template used when deploying Kiam with Helm).

tyrken:
Rebased & chart version bumped.  Added changelog entry but you'll have to fill in the date when you release.

rhysemmas:
Thanks @tyrken - as an aside, the spaces that were removed in your last commit (your text editor may have done this automatically) are markdown syntax for newlines, so the changelog will render differently now. I will put these back in when the changelog is updated with the date of the chart release

tyrken:
Thanks & sorry about the spaces!

PR #337, opened 2019-12-04
enable gRPC keepalive to detect dead TCP connections (#217)

mechpen:
When a TCP connection dies, i.e. no more packet from peer, gRPC
does not close the connection by default.  We need to enable gRPC
keepalive to detect and close dead TCP connections.


dbyron0:
What are the chances of releasing a new version with these changes?

PR #346, opened 2019-12-20
Constrain the regional endpoint resolver to STS service

### LINKED ISSUE #345: NoCredentialProviders/UnknownOperationException when using --region
In some (all?) regions, if you use the --region flag, kiam-server will fail to get credentials from the EC2 Metadata server to make the STS call.

You'll see errors like the following:
```
{"generation.metadata":0,"level":"error","msg":"error warming credentials: NoCredentialProviders: no valid providers in chain\ncaused by: EnvAccessKeyNotFound: failed to find credentials in the environment.\nSharedCredsLoad: failed to load profile, .\nEC2RoleRequestError: no EC2 instance role found\ncaused by
: EC2MetadataError: failed to make EC2Metadata request\ncaused by: \u003cUnknownOperationException/\u003e\n...
```

The reason is the --region flag defines an endpoint resolver, but doesn't constrain the resolver to the sts service. Fix is incoming.

cjbradfield:
Fixes #345

### DIFF
--- a/pkg/aws/sts/gateway.go
+++ b/pkg/aws/sts/gateway.go
@@ -38,7 +38,11 @@ type regionalResolver struct {
 }
 
 func (r *regionalResolver) EndpointFor(svc, region string, opts ...func(*endpoints.Options)) (endpoints.ResolvedEndpoint, error) {
-	return r.endpoint, nil
+	if svc == "sts" {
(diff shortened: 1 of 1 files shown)


acaire:
@Joseph-Irving Any possibility of a release please? 🙏

PR #355, opened 2020-01-10
update cert-manager docs for TLS config cert-manager v0.11.0+

cheeseandcereal:
cert-manager [version 0.11.0](https://cert-manager.io/docs/installation/upgrading/upgrading-0.10-0.11/) introduced a breaking api change to the deployment spec for all of its CRDs.

This PR introduces the changes needed in the TLS cert-manager docs to support cert-manager 0.11.0 or later.

Note the following breaking changes were required:
* Change the api version to `cert-manager.io/v1alpha2` which is the only supported api version starting with cert-manager 0.11.0
* Add `usages` to the certificate specs for the CA and server/agent certs. Without this, cert-manager now defaults the cert's `X509v3 Extended Key Usage` to be ONLY `TLS Web Server Authentication` by default, which breaks the CA certificate with the golang x509 verification library in gRPC

### DIFF
--- a/docs/TLS.md
+++ b/docs/TLS.md
@@ -61,10 +61,10 @@ kubectl create secret generic kiam-agent-tls -n kube-system \
 
 ## Cert manager
 
-You can use `cert-manager` to create a selfSigned issuer to create a CA and ca issuer for creating the required certs using that CA:
+You can use `cert-manager` to create a selfSigned issuer to create a CA and ca issuer for creating the required certs using that CA (note the following is only compatible with cert-manager version 0.11.0 or later):
 
-```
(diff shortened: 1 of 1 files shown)


mikesplain:
For others that may come across this in the future, the below statement especially applies to cert-manager v0.12.0 where this behavior is defaulted in https://github.com/jetstack/cert-manager/pull/2351

>Add usages to the certificate specs for the CA and server/agent certs. Without this, cert-manager now defaults the cert's X509v3 Extended Key Usage to be ONLY TLS Web Server Authentication by default, which breaks the CA certificate with the golang x509 verification library in gRPC

Thanks so much for documenting this @cheeseandcereal... totally missed it.

PR #360, opened 2020-01-16
Add support for external ID annotation

### LINKED ISSUE #236: External_ID Support for Cross account IAM Role

randomvariable:
Fixes #236 
Rebased version of #255

Adds support for external ID annotation. Instructions in updated README.

Note that https://github.com/patrickmn/go-cache can only work off a single key, hence this PR uses a pipe concatenation of the role and external ID as the cache key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,19 @@ metadata:
     iam.amazonaws.com/role: reportingdb-reader
 ```
 
+kiam also supports the use of an external id annotation added to the `Pod` which
+maybe used to avoid [confused deputy scenarios in cross-organisation role assumption](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html),
+for example:
+
+```yaml
+kind: Pod
+metadata:
+  name: foo
+  namespace: cross-org-iam-example
+  annotations:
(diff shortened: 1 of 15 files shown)


pingles:
I'll try and see if I can find some time to take a look through.

Appreciate the cache only supports a string key, I'd like to see if there's a way we can contain/separate the external ID stuff a little better to avoid string separation/concatenation across different parts of the code.

It'd be useful to see an example in pkg/aws/sts/arn_resolver_test.go that shows what happens with something that contains an externalID.

`STSGateway` now has quite a few parameters on `Issue` that would probably be nice to capture into a struct (role, session etc.).

Does that make sense? I'm happy to give it a go in Emacs. Alternatively, if you wanna try some remote pairing some time I can see if I can squeeze next week a little. Not done that before, would be fun!

randomvariable:
Happy to complete the solution, but probably worth pairing initially to come up with a design that satisfies everyone. Not sure about time at the moment, I'll get back to you.

randomvariable:
Rallying in #399

PR #361, opened 2020-01-16
Helm - Add possibility to configure agent/server livenessProbe and readinessProbe

caiohasouza:


helm/kiam/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 5.5.0
+version: 5.5.1
```
rhysemmas: Hey @caiohasouza, thanks for contributing! 

My only suggestion is that please can we bump the minor version here, rather than patch, as a backwards-compatible functionality change is being made

caiohasouza: Hello @rhysemmas , right, and i need change something in my commits? Or this change will be done for you?

rhysemmas: You'll need to update your commit so that the minor chart version is bumped instead (`5.6.0` instead of `5.5.1`), as when this PR is merged our CI will package a new chart using the version number included in the chart.

caiohasouza: Ok, done, could you check please?

PR #364, opened 2020-01-18
Add metrics for TLS certificate expiration

### LINKED ISSUE #362: Export Prometheus metric for TLS certificate expiration time
Server and agent should export a gauge set to the expiry time (seconds after unix epoch) of their cert.

abursavich:
Fixes #362

### DIFF
--- a/pkg/server/tls.go
+++ b/pkg/server/tls.go
@@ -1,0 +1,110 @@
+package server
+
+import (
+	"crypto/tls"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"sync"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	clientTLSMetrics = newTLSMetrics("client")
+	serverTLSMetrics = newTLSMetrics("server")
+)
+
+type tlsMetrics struct {
+	registerOnce sync.Once
+	verifyError  prometheus.Gauge
+	expiration   prometheus.Gauge
+}
+
+func newTLSMetrics(subsystem string) *tlsMetrics {
+	return &tlsMetrics{
+		verifyError: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: "grpc",
+			Subsystem: subsystem,
+			Name:      "tls_certificate_verify_error",
+			Help:      "Indicates if there was an error verifying the latest gRPC " + subsystem + " TLS certificate and its expiration.",
+		}),
+		expiration: prometheus.NewGauge(prometheus.GaugeOpts{
+			Namespace: "grpc",
+			Subsystem: subsystem,
+			Name:      "tls_certificate_expiry_seconds",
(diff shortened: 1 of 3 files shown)


pkg/server/tls.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func newLazyGauge(opts prometheus.GaugeOpts) *lazyGauge {
+	return &lazyGauge{metric: prometheus.NewGauge(opts)}
+}
+
+func (g *lazyGauge) Set(v float64) {
+	g.init.Do(func() { prometheus.MustRegister(g.metric) })
+	g.metric.Set(v)
+}
```
Joseph-Irving: what is the reason for creating this wrapper around gauge? I'd rather keep things simple and readable instead of creating a custom struct for the sake of saving a few lines here and there

pkg/server/tls.go
```diff
@@ -0,0 +1,85 @@
...
+
+func loadCerts(certFile, keyFile, caFile string) (tls.Certificate, *x509.CertPool, error) {
+	cert, err := tls.LoadX509KeyPair(certFile, keyFile)
+	if err != nil {
+		return tls.Certificate{}, nil, err
+	}
+	ca, err := ioutil.ReadFile(caFile)
+	if err != nil {
+		return tls.Certificate{}, nil, err
+	}
```
Joseph-Irving: we've lost some detail in the returned error here, previously we specified things like `error loading keypair` in the returned error message, which can just make debugging it a little easier for people.

pkg/server/tls.go
```diff
@@ -0,0 +1,85 @@
...
+func verify(cert tls.Certificate, pool *x509.CertPool, usage x509.ExtKeyUsage) (chains [][]*x509.Certificate, err error) {
+	x509Cert := cert.Leaf
+	if x509Cert == nil {
+		x509Cert, err = x509.ParseCertificate(cert.Certificate[0])
+	}
+	return x509Cert.Verify(x509.VerifyOptions{
+		Roots:     pool,
+		KeyUsages: []x509.ExtKeyUsage{usage},
+	})
+}
```
Joseph-Irving: we didn't do any key usage verification previously so this would potentially be a breaking change for users, so this would either need to be optionally enabled or it would have to a wait for a major version change.

Joseph-Irving:
Sorry for taking a while to get around to looking at this, I like the feature, I'm sure people would get use out of it. I've left a few comments.

pkg/server/tls.go
abursavich: I only wanted to register the metrics when they are used (e.g. I didn't want the client metric to exist on the server and vice versa).

pkg/server/tls.go
abursavich: Done.

pkg/server/tls.go
abursavich: No longer required, but tracked in a separate metric.

abursavich:
Please take another look, thanks!

pkg/server/tls.go
abursavich: I added more metrics that are symmetric (server and client), so I moved them into a struct that self-registers.

Issue #358, opened 2020-01-14
kiam-agent random errors on node startup


caiohasouza:
Hello,

I'm have some random errors when the new nodes are added to cluster, randomly the kiam agent fail on start with error

> {"level":"fatal","msg":"error creating server gateway: error dialing grpc server: context deadline exceeded","time":"2020-01-14T17:53:00Z"}

After some restarts the kiam-agent works normally, this happen randomly but only in some nodes:
> kiam-agent-4rxlg                           1/1     Running   0          2m37s
kiam-agent-4whq4                           1/1     Running   1          2m46s

This behavior is normal? I need fix/change some parameter?

Thank you!

nirnanaaa:
we're experiencing the exact same behavior when rolling out new kiam servers. What we have seen is that for most agents the grpc connection is updated to those new servers. but for some reason sometimes it just stays as is.

caiohasouza:
When i enabled the GRPC debub i see this error:

> grpc: failed dns SRV record lookup due to lookup _grpclb._tcp.kiam-server on 100.64.0.10:53: no such host.

So i my case the problem is with dns resolution name, i created a PR to be possible add a kiam-agent initContainer (https://github.com/uswitch/kiam/pull/367). With initContainer it will possible check the dns resolution before kiam-agent start:

>   initContainers:
    - name: check-dns
      image: busybox:1.31
      imagePullPolicy: IfNotPresent
      command: ['sh', '-c', 'until nslookup kiam-server; do sleep 1; done;']
      resources:
        requests:
          cpu: 20m
          memory: 16Mi
        limits:
          cpu: 20m
          memory: 16Mi

I tested in my local environment and the problem was fixed.

nirnanaaa:
I think such a change only works against the theories behind k8s mainly the exponential backoff and fail fast theorem. The agent should just die in case it gets to a state where it can't recover from, right ?

Joseph-Irving:
The problem would appear to be that the Kiam node is starting on your nodes before DNS is working properly on them (maybe CNI still initialising something like that), in that case it's perfectly reasonable for Kiam to restart. 

Is the kiam agent restarting causing you application level issues?
If so it might be worth implementing a fix for your setup to prevent this otherwise it's probably fine to leave it as is.

nirnanaaa:
For us at least this is happening not on startup but just Midway through.

caiohasouza:
[...]

PR #367, opened 2020-01-20
Helm Chart 5.5.2 - Add possibility to configure agent/server initContainers

caiohasouza:
Fix https://github.com/uswitch/kiam/issues/358

### DIFF
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 5.5.2
+version: 5.6.1
 appVersion: 3.5
 description: Integrate AWS IAM with Kubernetes
 keywords:
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 5.5.0
+version: 5.5.2
 appVersion: 3.5
 description: Integrate AWS IAM with Kubernetes
 keywords:
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 5.6.1
+version: 5.7.0
 appVersion: 3.5
 description: Integrate AWS IAM with Kubernetes
 keywords:
--- a/helm/kiam/values.yaml
+++ b/helm/kiam/values.yaml
@@ -133,6 +133,23 @@ agent:
     #   hostPath: /usr/share/ca-certificates
     #   readOnly: true
 
+  ## Agent initContainers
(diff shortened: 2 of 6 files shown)


caiohasouza:
Someone?

helm/kiam/values.yaml
```diff
@@ -133,6 +133,23 @@ agent:
...
+    #   imagePullPolicy: IfNotPresent
+    #   command: ['sh', '-c', 'until nslookup kiam-server; do sleep 1; done;']
+    #   resources:
+    #     requests:
+    #       cpu: 20m
+    #       memory: 16Mi
+    #     limits:
+    #       cpu: 20m
+    #       memory: 16Mi
+
```
Joseph-Irving: I think having the option to configure init containers if you desire is great, but I wouldn't want this example included in the repo, it's very specific to a problem you were experiencing in your cluster

caiohasouza: No problem, i removed!

helm/kiam/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 5.5.0
+version: 5.6.1
```
rhysemmas: Please bump the minor version, rather than patch, as we are making a backwards compatible functionality change

helm/kiam/values.yaml
```diff
@@ -271,6 +276,23 @@ server:
   ##
   extraHostPathMounts: []
 
+  ## Server initContainers
+  ## Ref: https://kubernetes.io/docs/concepts/workloads/pods/init-containers/
+  ##
+  initContainers: []
+    ## Check kiam-server dns hostname
```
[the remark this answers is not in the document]
caiohasouza: sure, done!

helm/kiam/Chart.yaml
caiohasouza: @rhysemmas sorry, i don't understand, currently the PR version is 5.6.1, what version you want that i configure?

rhysemmas: Hey @caiohasouza, Helm charts use semantic versioning to indicate major/minor/patch updates to the functionality of the chart (see https://semver.org/ ). So in this case your PR will introduce a backwards-compatible change, which by semver standards means we should increase the minor version of the chart. The chart version is currently `5.6.1`, so in your PR the chart version should be bumped to `5.7.0`

caiohasouza: Hi @rhysemmas , thank you for you explanation. Ok, version changed!

rhysemmas:
Thanks for this @caiohasouza! Could you please resolve the conflict (your chart version bump is comparing to an older git commit than the current master) and we'll get this merged in

caiohasouza:
@rhysemmas done, could you check please?

PR #372, opened 2020-01-30
Add missing sslCertHostPath as allowedHostPath in Helm Chart server PSP

phyrog:
When using PSPs, the current helm chart does not automatically add the `sslCertHostPath` as an `allowedHostPath`, which causes the server to be unable to start.

This also fixes the (I think) misplaced `range` in the PSP.


rhysemmas:
Thanks for contributing @phyrog! Great catch, and yes that range of the `extraHostPathMounts` is now in the right place

PR #374, opened 2020-01-31
Remove unnecessary note about older cert-manager versions

ernetas:
Line 64 and the whole example configuration already imposes a requirement of at least v0.11.0 since https://github.com/uswitch/kiam/pull/355, so the removed line is no longer relevant


Joseph-Irving:
thanks

PR #380, opened 2020-03-02
Refactor unnecessary usage of channels

cPu1:
Some handlers create a new local channel, send to it in the `backoff.Retry` handler, and receive from it after the `backoff.Retry` call returns with a non-nil error. This is all sequential code that needs no synchronization (and its overhead), and can be expressed without using any synchronization mechanism like channels.

This changelist removes the usage of channels where it's not required and refactors it to use local assignment.


Joseph-Irving:
@pingles this looks good to me, was there any reason we used channels here or is it just a leftover?

PR #381, opened 2020-03-03
Support IMDS v2

### LINKED ISSUE #359: Unable to access new AWS metadata api
Amazon recently [updated their metadata api.](https://aws.amazon.com/about-aws/whats-new/2019/11/announcing-updates-amazon-ec2-instance-metadata-service/)  When I attempt to follow the basic instructions to retrieve metadata as [described here](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html), kiam rejects the request with:
`request blocked by whitelist-route-regexp "^$": /latest/api/token`

rbvigilante:
fixes #359 

- Prevent the reverse proxy adding an X-Forwarded-For header
- Allow PUT requests to `*/api/token` without consulting the API whitelist
- Authenticate handing out roles by checking the session against the metadata API

### DIFF
--- a/pkg/aws/metadata/handler_proxy.go
+++ b/pkg/aws/metadata/handler_proxy.go
@@ -27,6 +27,8 @@ type proxyHandler struct {
 	whitelistRouteRegexp *regexp.Regexp
 }
 
+var tokenRouteRegexp = regexp.MustCompile("^/?[^/]+/api/token$")
+
 func (p *proxyHandler) Install(router *mux.Router) {
 	router.PathPrefix("/").Handler(adapt(withMeter("proxy", p)))
 }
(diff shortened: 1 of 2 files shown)


Joseph-Irving:
the tests are failing as you're using a http method (request.Clone) not found in go1.12, I've updated the go version, so if you rebase with master it should pass

rbvigilante:
Ah, sorry. Rebase done.

Joseph-Irving:
Hey, I've been trying to look over this, the first two commits seem pretty self-explanatory. It's the third one that I keep getting stuck on, can you give some more detail about what you're trying to achieve here? Is it actually needed for this to work or is it an extra set of security you're adding? If it's the latter perhaps we should split this up.

rbvigilante:
The reasoning behind the third commit went something like this:

If I'm on an AWS instance (without Kubernetes or Kiam), I can now (successfully) get credentials out of the metadata API in two ways:
 - I can just hit `169.254.169.254/latest/meta-data/iam/security-credentials/<role>` with no headers, and get the credentials through IMDSv1 (unless this has been explicitly disabled)
 - I can get a token, then hit the same URL _with a session token header_ to get the credentials through IMDSv2

I'm pretty sure the first two commits here (whitelisting the token endpoint and removing the `x_forwarded_for` header) _will_ allow IMDSv2 to function more or less as intended for every endpoint Kiam isn't taking over. For the ones it _does_ take over (`/iam/*`), though, we'd be potentially accepting a session token and completely ignoring it (and handing out a role regardless).

The third commit is my attempt to respect the session control aspect of IMDSv2 without needing to keep track of sessions within Kiam. Totally happy to remove it and deal with it later if you like.

Joseph-Irving:
Right ok, I think I follow, so you're having kiam check the token before handing out a role to the requester as an additional security layer.
That seems like a nice thing to have but I would be keen to have this separated out of this PR and potentially put behind a feature flag, I'm worried about this interacting with certain users' setups in a strange way and causing kiam to totally stop working for them.

rbvigilante:
Sure. I'll put it into another branch.

rbvigilante:
Done

PR #384, opened 2020-03-07
Return credentials with trailing slash

### LINKED ISSUE #383: KIAM doesn't correctly handle trailing slashes on security-credentials requests
Real metadata service:
```
curl -i http://169.254.169.254/latest/meta-data/iam/security-credentials/role_name/
HTTP/1.0 200 OK
[snip]

curl -i http://169.254.169.254/latest/meta-data/iam/security-credentials/role_name
HTTP/1.0 200 OK
[snip]
```

KIAM:
```
curl -i http://169.254.169.254/latest/meta-data/iam/security-credentials/role_name/
HTTP/1.1 500 Internal Server Error
Content-Type: text/plain; charset=utf-8
X-Content-Type-Options: nosniff
Date: Fri, 06 Mar 2020 23:34:45 GMT
Content-Length: 48

error fetching credentials: forbidden by policy

curl -i http://169.254.169.254/latest/meta-data/iam/security-credentials/role_name
HTTP/1.0 200 OK
[snip]
```

We tried changing the `--whitelist-route-regexp` value but that had no effect.

sporkmonger:
Fixes #383.

Adds a second handler with a trailing slash. I initially tried doing this with a single handler but it broke all the tests.

Opted not to use the redirect method used elsewhere because, in this case, the version without the trailing slash is canonical but clients affected by this issue aren't guaranteed to follow redirects and the real metadata service returns a 200 for both.

### DIFF
--- a/pkg/aws/metadata/handler_credentials.go
+++ b/pkg/aws/metadata/handler_credentials.go
@@ -17,13 +17,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
+
 	"github.com/cenkalti/backoff"
 	"github.com/gorilla/mux"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/uswitch/kiam/pkg/aws/sts"
 	"github.com/uswitch/kiam/pkg/server"
 	"github.com/uswitch/kiam/pkg/statsd"
-	"net/http"
 )
 
 type credentialsHandler struct {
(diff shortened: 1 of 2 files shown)


pingles:
It'd be good to verify this behaviour across a few instance generations (I know historically there's been some variation there) but definitely looks like something to fix. Do you remember which you were using to see the behaviour above?

Afterwards could you take a look at updating to bring in line with other v4 changes, it shouldn't be too tricky.

PR #394, opened 2020-04-20
Continue startup if the statsd endpoint can't be set up

rbvigilante:
We noticed that on startup Kiam, if for whatever reason it can't resolve its statsd endpoint, exits with a fatal error. Since it's _possible_ for Kiam to run without a statsd endpoint, how would you feel about continuïng startup in this case? 

No real pressure (I can understand why you might like to have it work either way), and also totally cool if this implementation isn't up to scratch for whatever reason, but I made a pull request because it seemed like a pretty small change.

Thank you!

### DIFF
--- a/cmd/kiam/opts.go
+++ b/cmd/kiam/opts.go
@@ -79,7 +79,9 @@ func (o telemetryOptions) start(ctx context.Context, identifier string) {
 	)
 
 	if err != nil {
-		log.Fatalf("Error initing statsd: %v", err)
+		log.Errorf("Error initing statsd: %v", err)
+		log.Error("Continuing startup with statsd disabled...")
+		statsd.New("", "", o.statsDInterval)
 	}
 
 	if o.prometheusListen != "" {


Joseph-Irving:
Unless I'm mistaken, this error would only occur if you're specifying the statsd url and kiam can't resolve it, as by default statsd is disabled. 
So If you're specifying the statsd url it does imply you specifically want statsd metrics for kiam, so in that scenario, if resolution fails but kiam starts up anyway I have a pod with no metrics for which I'll need to go and manually restart at some point to get metrics back.

rbvigilante:
That's correct - you'd need to be specifying the endpoint, then having it break. This can also happen dynamically, while a cluster is running - if the statsd endpoint becomes unresolvable, any new nodes that are scaled into your cluster won't have Kiam, which will mean pods (probably) won't be able to run on them.

It's a fairly far-out edge case - like I said, fine for you to just close this off if you prefer.

pingles:
Thanks for sharing this but in v4 to tidy it up we've removed StatsD support #429.

rbvigilante:
OK, no problem. Will the metrics that are currently only available over statsd be made available through Prometheus?

PR #399, opened 2020-04-29
Add support for external ID annotation (follow up)

etwillbefine:
Hey there,

I came across the issue https://github.com/uswitch/kiam/issues/236 and the related PR https://github.com/uswitch/kiam/pull/360. It looks the the PR got stuck a little bit. 

Hopefully this addresses https://github.com/uswitch/kiam/pull/360#issuecomment-575080120 by introducing `STSGatewayRequest ` and move the external ID out of the role. 

Im not fully sure whether it really replaces the need of the concatenation. I added two more tests to announce a new pod with an external ID set. Tested it on my own cluster using this [docker image](https://hub.docker.com/r/gocidocker/kiam-with-external-id). Access will be denied if no or wrong annotation is present and credentials can be retrieved once the correct external-id annotation is added. Refreshing credentials didnt worked in the first try, addressed it in https://github.com/uswitch/kiam/pull/399/commits/cb8c1197f132be731fbea98c397eeb51059dddff

As AWS itself does not treat the External ID secret I don't want to introduce new secret for now as suggested [here](https://github.com/uswitch/kiam/issues/236#issuecomment-488884021). I'm also not sure about any performance implications or something else this could have. https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

@pingles it would be great if you could have a look.
/cc @randomvariable


randomvariable:
Happy for you to finish this off actually. I don't have the time right now to work on it.

etwillbefine:
I would really appreciate some feedback to finally get this merged. Anyone can help? Maybe @pingles ?

stefansedich:
@etwillbefine was there ever any more movement on this? I myself started looking at adding support for session names but soon hit a roadblock in that it would require some changes around how caching is done today in that the cache key would need to add the session name, I assume the same would be true for the external-id.

@pingles what was the status of this PR? I wonder if we could combine efforts to get session-name and external-id done at the same time including the required changes to the cache to handle caching unique credentials based on role/external-id/session-name.

etwillbefine:
So far there hasn't been any progress. I am using the mentioned Docker Image in latest Release successfully for several months now. If there are any additional Tests needed I'm happy to free up some time to get this merged.

pingles:
Thanks for the contribution, and sorry for not taking a look earlier. I run out of time to help contribute much these days.

I've got a few hours blocked tomorrow morning so I'll try and take a look then.

pingles:
Ok, so having had a scan through, you hadn't made a change to the caching mechanism in the same way that @randomvariable did. I think that's ok (the caching stuff is a bit of a mess at the moment so would love to clean up and address the session naming issue).

I don't think it's necessary, @randomvariable do you recall a reason why you chose to do it that way? Do AWS connect secret keys to access key and external ID in some way that this would mess up auditing?

If that's all ok then I'd rather implement this way and then try to resolve the caching and session naming stuff separately.

@uswitch/cloud any comments, would trust your opinions also 😄

randomvariable:
Yes, the access key, secret key, session token tuple is tied directly to the parameters to sts.assumerole call. Absolutely need to disambiguate pods with different or no external IDs to resolve the confused deputy problem.

pingles:
Awesome. I'll pull the branch and take a look through. I can see this continues on from your work so think I've misunderstood the implementation from my earlier comment.

pingles:
So looking through the code it looks like the key change is to cache with both the role and external ID.

Adding a `cacheKey(role, externalID)` func enough and have that do the formatting?

randomvariable:
Yeah, I think so. I did it the way I did because I was guessing at the communications between the various components.

pingles:
I dusted off Emacs and implemented it :-)

I'll push my commits up when I'm back at my desk with my Yubikey and then be glad to get some other opinions. I'm spotting plenty of other things that could be improved!

pingles:
So I pushed my change to https://github.com/uswitch/kiam/tree/etwillbefine-external-id-4 but have just noticed `service.proto` was changed to add `ExternalID` to `Role`. That shouldn't be necessary: all of the management of the Pod data should be able to be handled by the Server internally.

I think it may mean reworking a few things. I'll try and take a look later today or over the weekend.

pingles:
Ok so this will turn out to be easier to do if we tidy up a bunch of other things (dropping the deprecated services).

I'm going to pick up the changes here in another branch, remove the old RPCs and see if I can add something for doing session naming too.

stefansedich:
@pingles that is looking good, I found when I tried to implement session name support would need changes to the cache item key, your changes will let me implement that once this is merged :smile:

pingles:
I screwed up the merge on GH having made the RPC change. I'll pick up the code from the branch here and do manually, hopefully some time over the next few days.

pingles:
@stefansedich I'm hoping to get some time over the next few days to merge some of the changes in here. I think it'll be easier to get these changes merged into main and then go from there?

In the meantime, do you want to add more info to #38 for how you think it could work, whether we'd support templates (naming based on Pod name for example) etc.

pingles:
@stefansedich That would be great, thanks!

I removed some of the deprecated RPCs in #426, and had started to make a few other changes in the v4-wip branch https://github.com/uswitch/kiam/tree/v4-wip. 

I was going to start work on pulling the changes from this into there and then take a look through the session naming. Does that help?

stefansedich:
@etwillbefine I am tackling this over in https://github.com/uswitch/kiam/pull/430 which :crossed_fingers: goes in soon!

pingles:
I've just merged @stefansedich PR so this is now in main. We'll try and get some of the other v4 things in (upgrade notes in particular) before we start doing some RCs but you can use a main sha to early test if you like.

PR #402, opened 2020-05-13
add retries around iptable rule removal

### LINKED ISSUE #396: iptables - should have retries in cases where it fails to update
The iptables implementation tries to add a rule when KIAM-Agent daemonset is started, if that fails (due to lock, or other issue), the container restarts, the command is retried and usually this is not a problem.

But in the case of shutdown, this can be problematic because the rule might get left behind after the daemonset is deleted.

In our use case, we have implemented a solution for "blue/green" of sort for KIAM in order to avoid the downtime when we rotate cert, and we do this by deploying a second full stack of KIAM (new certs, new agents, new server, new service), and then delete the old stack of KIAM which removes it's rule and cuts over the traffic to the new stack -> cert rotated and 0 downtime.

However this gets problematic in cases where the IPTable rule is not properly removed by the deferred call (either a fatal exit, or iptables lock fails).

We should evaluate either having retries around `iptables.Remove()` or possibly use the binary's `--wait` / `--wait-interval` if the `iptables-go` pkg allows it, which will let the underlying iptables binary do the waiting for the lock.

Also we should make sure in cases of fatal exit we also remove the iptables rule - Ref PR: #300 

Will it be possible to address this? I can send a PR if you agree with this.

eytan-avisror:
Fixes #396 

This adds basic retries around removing the IPTable rule for kiam-agent when the pod is terminated.

Given the process is exiting anyway, we cannot really 'guarantee' removal, but these retries will make it highly unlikely that a rule is left behind.

The call to `Remove()` gets deferred regardless of error anyway, so we might as well retry internally.

https://github.com/uswitch/kiam/blob/6e07571932910d71a8d99a7b4dd0325e09baf9b4/cmd/kiam/agent.go#L68-L70

### DIFF
--- a/cmd/kiam/iptables.go
+++ b/cmd/kiam/iptables.go
@@ -79,7 +79,7 @@ func (r *rules) Remove() error {
 			break
 		}
 		if err := ipt.Delete("nat", "PREROUTING", r.ruleSpec()...); err == nil {
-			log.Info("iptables rule was removed", err.Error())
+			log.Info("iptables rule was successfully removed")
 			break
 		}
 		log.Warnf("failed to remove iptables rule, will retry: %s", err.Error())
--- a/cmd/kiam/iptables.go
+++ b/cmd/kiam/iptables.go
@@ -16,8 +16,10 @@ package main
 import (
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/coreos/go-iptables/iptables"
+	log "github.com/sirupsen/logrus"
 )
 
 type rules struct {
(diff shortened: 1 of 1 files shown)


shrinandj:
This is awesome! Thanks for doing this ..

cmd/kiam/iptables.go
```diff
@@ -59,13 +61,32 @@ func (r *rules) ruleSpec() []string {
...
 
-	return ipt.Delete("nat", "PREROUTING", r.ruleSpec()...)
+	var attempt int
+	for {
+		if attempt >= maxAttempts {
+			log.Errorf("failed to remove iptables rule, retries exhausted: %s", err.Error())
+			break
+		}
+		if err := ipt.Delete("nat", "PREROUTING", r.ruleSpec()...); err == nil {
+			log.Info("iptables rule was removed", err.Error())
```
Joseph-Irving, line 82: why have the error here?  we've already checked that err is nil

eytan-avisror: Removed it, thanks

PR #405, opened 2020-05-29
fix server affinity

msvechla:
Hi, 

Currently, there is an issue with the server deployment and affinity. When an affinity is set we get the following error:

> Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.affinity.podAntiAffinity): unknown field "nodeAffinity" in io.k8s.api.core.v1.PodAntiAffinity

This is due to wrong indentation on the affinity spec. This PR fixes this issue.
Let me know if you need anything else.


msvechla:
Any updates here? This is rendering the chart unusable when affinity is used.

PR #411, opened 2020-07-13
Update changelog for v3.6

rhysemmas:


waiariki-koia:
Do we need to try fix codecov?

Joseph-Irving:
I think codecov is broken @waiariki-koia, we should fix that separately

PR #414, opened 2020-08-03
Add documentation about AWS cross-account IAM use case

123BLiN:
I spent a couple of days trying to get Grafana pod to work with multiple CloudWatch datasources as well as Prometheus to discover targets in multiple AWS accounts. So maybe this small update will save time to others.

### DIFF
--- a/docs/IAM.md
+++ b/docs/IAM.md
@@ -172,3 +172,40 @@ The cluster node EC2 instance role will need to have `sts:AssumeRole` permission
 for this role and for there to be an entry in the Server role's trust policy.
 Application roles will need to have a trust policy entry for this role, instead
 of the cluster node role as noted above.
+
+## Cross-Account IAM Roles
+
+Cross-account access to resources in AWS with Kiam can be configured in multiple ways, simple setup requires specifiyng full role ARN (including account ID) from another account in the `iam.amazonaws.com/role` pod annotation. Same rules are valid for this IAM role - it needs to have `sts:AssumeRole` permissions from the Kiam server role.
+
+If you need to provide cross-account access to multiple AWS accounts at once to single pod you will need to have additional IAM role in source AWS account with permissions to assume target IAM roles.
+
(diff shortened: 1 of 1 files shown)


pingles:
I think it's good to document this kind of scenario (I seem to remember it's come up before) where people have wanted to be able to assume more than one role to a pod.

Having said that, I'm not sure I follow the example policies. Maybe it's worth documenting the example with a picture to help illustrate.

In the README there's a link to a video that @jbeda recorded and he drew a really good sketch of how roles and policies came together, using that as a starting point would be cool.

docs/IAM.md
```diff
@@ -172,3 +172,40 @@ The cluster node EC2 instance role will need to have `sts:AssumeRole` permission
...
 Application roles will need to have a trust policy entry for this role, instead
 of the cluster node role as noted above.
+
+## Cross-Account IAM Roles
+
+Cross-account access to resources in AWS with Kiam can be configured in multiple ways, simple setup requires specifiyng full role ARN (including account ID) from another account in the `iam.amazonaws.com/role` pod annotation. Same rules are valid for this IAM role - it needs to have `sts:AssumeRole` permissions from the Kiam server role.
+
+If you need to provide cross-account access to multiple AWS accounts at once to single pod you will need to have additional IAM role in source AWS account with permissions to assume target IAM roles.
+
+Imagine next case scenario - service in the source account `AWS-source` requires access to S3 buckets in target accounts `AWS-target1` and `AWS-target2`. You are not allowed to assing multiple roles to single Pod with Kiam so you have to create service IAM role in the `AWS-source` account and assing it to the pod, this role must have permissions to assume roles in other accounts, for example with such inline policy:
```
sylr, line 182:
```suggestion
Imagine next case scenario - service in the source account `AWS-source` requires access to S3 buckets in target accounts `AWS-target1` and `AWS-target2`. You are not allowed to pass multiple roles to single Pod with Kiam so you have to create service IAM role in the `AWS-source` account and passing it to the pod, this role must have permissions to assume roles in other accounts, for example with such inline policy:
```

sylr (changes requested):

PR #415, opened 2020-08-05
Allow not mounting SSL certs from host

velothump:
We're looking to migrate our cluster from CoreOS to FCOS, and they host SSL path differs between the host. This PR will enable us to use our own Kiam image with the certs already added to facilitate the upgrade.


Joseph-Irving:
lgtm, please can you rebase with master and bump version number accordingly

velothump:
@Joseph-Irving Thanks, rebased now :)

PR #417, opened 2020-08-09
Helm chart 3.6

leosunmo:
This updates the Helm chart's default values to use kiam v3.6.

Reading through the v3.6 release it didn't look like any modifications needed to be done to the Helm charts values other than bumping image tags.

I made the default values readable in raw text form.

I also backfilled the changelog as the gaps were annoying me!

Make sure to replace placeholders in changelog before merging. Maintainers, feel free to push directly to the PR with any changes.

### DIFF
--- a/helm/kiam/CHANGELOG.md
+++ b/helm/kiam/CHANGELOG.md
@@ -1,5 +1,42 @@
 # Helm Chart Changelog
 
+## 5.9.0
+REPLACEME 2020
+
+Notable changes:
+* [#406](https://github.com/uswitch/kiam/pull/REPLACEME) Update kiam release from 3.5 to 3.6.
+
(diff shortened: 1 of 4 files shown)


helm/kiam/CHANGELOG.md
```diff
@@ -1,5 +1,42 @@
 # Helm Chart Changelog
 
+## 5.9.0
+REPLACEME 2020
```
Joseph-Irving, line 4: this all looks good thanks, just need to remove this line

leosunmo: Awesome, thanks for looking! 

The idea was that we'd replace that when merging :)
If you want I can change it now or you can change as you merge.

PR #423, opened 2020-09-10
Adding maxUnavailable for daemonsets

gregsidelinger:


helm/kiam/templates/agent-daemonset.yaml
```diff
@@ -167,4 +167,8 @@ spec:
       {{- end }}
   updateStrategy:
     type: {{ .Values.agent.updateStrategy }}
+{{- if eq .Values.agent.updateStrategy "rollingUpdate" }}
```
stefansedich, line 170: Should this be `RollingUpdate`?

helm/kiam/templates/server-daemonset.yaml
```diff
@@ -169,4 +169,8 @@ spec:
       {{- end }}
   updateStrategy:
     type: {{ .Values.server.updateStrategy }}
+{{- if eq .Values.server.updateStrategy "rollingUpdate" }}
```
stefansedich, line 172: Should this be `RollingUpdate`?

stefansedich:
Also don't forget to bump the charts minor version in https://github.com/uswitch/kiam/blob/master/helm/kiam/Chart.yaml

PR #428, opened 2020-10-16
Refactoring credentials cache to aid in easier support for session-name and external-id

stefansedich:
This PR begins work towards supporting external-id and session-name by refactoring the caching layer to easier support things moving forward.

It introduces a struct `CredentialsIdentity` that will contain the role/session-name/extrernal-id and a function to generate the cache key. It also changes the cache to hold the `CachedCredentials` item so that expiry can get the identity it needs to refresh the role when required.

With this merged I will move onto the implementation for external-id and session-name support.

### DIFF
--- a/pkg/aws/sts/cache.go
+++ b/pkg/aws/sts/cache.go
@@ -28,15 +28,19 @@ type credentialsCache struct {
 	arnResolver     ARNResolver
 	baseARN         string
 	cache           *cache.Cache
-	expiring        chan *RoleCredentials
+	expiring        chan *CachedCredentials
 	sessionName     string
 	sessionDuration time.Duration
 	cacheTTL        time.Duration
 	gateway         STSGateway
 }
 
-type RoleCredentials struct {
-	Role        string
+type CredentialsIdentity struct {
+	Role string
+}
+
+type CachedCredentials struct {
+	Identity    *CredentialsIdentity
 	Credentials *Credentials
 }
 
@@ -53,7 +57,7 @@ func DefaultCache(
 ) *credentialsCache {
 	c := &credentialsCache{
 		arnResolver:     resolver,
-		expiring:        make(chan *RoleCredentials, 1),
+		expiring:        make(chan *CachedCredentials, 1),
 		sessionName:     fmt.Sprintf("kiam-%s", sessionName),
 		sessionDuration: sessionDuration,
 		cacheTTL:        sessionDuration - sessionRefresh,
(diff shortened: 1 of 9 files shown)


pkg/aws/sts/cache.go
```diff
@@ -77,70 +81,82 @@ func DefaultCache(
...
+		c.cache.Delete(identity.Role)
 		return nil, err
 	}
 
-	return val.(*Credentials), nil
+	cachedCreds := val.(*CachedCredentials)
+	return cachedCreds.Credentials, nil
+}
+
+func (c *credentialsCache) cacheKey(identity *CredentialsIdentity) string {
```
pingles, line 160: Thanks for adding. I'd change this so that perhaps `String()` on `CredentialsIdentity` generates the cache key, that way all the parts can be encapsulated away from `credentialsCache`. What do you think @stefansedich?

stefansedich: That sounds entirely reasonable, just made the change.

pkg/aws/sts/cache.go
```diff
@@ -77,70 +81,81 @@ func DefaultCache(
...
+	item, found := c.cache.Get(identity.String())
 
 	if found {
 		future, _ := item.(*future.Future)
 		val, err := future.Get(ctx)
 
 		if err != nil {
 			logger.Errorf("error retrieving credentials in cache from future: %s. will delete", err.Error())
-			c.cache.Delete(role)
+			c.cache.Delete(identity.Role)
```
pingles, line 117: `identity.Role` will need to be `identity.String()` to match when it's set and retrieved.

pkg/aws/sts/cache.go
```diff
@@ -77,70 +81,81 @@ func DefaultCache(
...
+		return cachedCreds, err
 	}
 	f := future.New(issue)
-	c.cache.Set(role, f, c.cacheTTL)
+	c.cache.Set(identity.String(), f, c.cacheTTL)
 
 	val, err := f.Get(ctx)
 	if err != nil {
-		c.cache.Delete(role)
+		c.cache.Delete(identity.Role)
```
pingles, line 151: sorry, another `identity.String()` key

pingles (changes requested):
Found a few places we'll need to use the new Identity `String()` func but otherwise looks nice to me, thanks for cleaning it up.

pkg/aws/sts/cache.go
stefansedich: Ouch my bad! I guess I had relied on tests a little much here, thanks for the careful eye. This should be fixed.

pkg/aws/sts/cache.go
stefansedich: Fixed.

pkg/aws/sts/cache.go
pingles: No worries. Yeah, absolutely should have had some tests to catch, sorry.

stefansedich: I can look at what coverage I can add as I get through the rest of the work.

pingles (approved):
Thanks!

pingles:
I’d love PRs adding more tests :-)

Thanks!

Issue #429, opened 2020-10-16
Deprecate StatsD for v4


pingles:
In the spirit of tidying code up, deleting StatsD support seems like a good idea given the tendency for using Prometheus with Kubernetes applications.

I'm keen to see what others think so opening this as an idea. Please comment / thumbs up or down to indicate preference.

PR #431, opened 2020-10-16
Delete statsd

pingles:
Inspired by #429 and other deprecation tidy-up, I thought this would be a good thing to consider.

I think it tidies up nicely.


pingles:
@Joseph-Irving what do you think?

Joseph-Irving (approved):
lgtm, Prometheus is the metrics standard these days, don't see any reason to carry this into v4.

Issue #368, opened 2020-01-23
kiam-server still accesses to STS global endpoint when setting --region


offzale:
Kiam-server is still accessing the global endpoint when setting the `--region` argument, instead of using the STS regional endpoint. Even setting both `--region` argument and `AWS_REGION` environment variable (as https://github.com/uswitch/kiam/issues/344 suggests) to the region value, `us-east-1`.

After limiting the egress to the internet (allowing access to AWS Metadata) and setting the necessary VPC Endpoints to access to AWS Services (including regional STS endpoint), we get this error in the server:

`"level":"error","msg":"error requesting credentials: RequestError: send request failed\ncaused by: Post https://sts.amazonaws.com/: dial tcp 54.239.29.25:443: connect: connection timed out"`

We would expect kiam-server to use the regional endpoint instead, `sts.us-east-1.amazonaws.com`. Is there any reason for kiam-server to still access the global endpoint? or maybe we are missing some configuration?

Kiam-server version: 3.5

pingles:
This definitely seems like a bug, not sure what's happening.

If someone is able to write some tests around it to try and prove the error and help fix I'd really appreciate. I'll try and take a look when I can but can't promise when.

pingles:
So having re-read the code, I wonder if the issue is actually the ordering of this:

https://github.com/uswitch/kiam/blob/master/pkg/aws/sts/gateway.go#L88

If `assumeRoleArn` is set, then Kiam creates a session to retrieve credentials for the server role, to then use ahead of all future calls.

Line 89 doesn't use the Region information when doing this, so it will end up defaulting to the global endpoint.

https://github.com/uswitch/kiam/blob/master/pkg/aws/sts/gateway.go#L89

Does that sound plausible given your setup @offzale?

@randomvariable @Joseph-Irving is there a reason it would need to use the global endpoint for retrieving server credentials, or can those also be forwarded to the local endpoint?

randomvariable:
Doesn't matter which endpoint is used as long as it's reachable. The credentials will always be globally scoped.

Hitting the global endpoint becomea an issue in Govcloud/Secret partitions though, so ideally would use the local endpoint.

pingles:
Awesome, thanks Naadir. I'll try and refactor it around, add some tests but sounds like this _may_ have been the cause of the noted behaviour.

pingles:
I've just pushed a change in that I believe should address this (and added tests to confirm the resolver is working correctly).

[...]

Issue #344, opened 2019-12-18
--region not handling sts endpoint as expected; still needs AWS_REGION environment variable


p-kossa:
I've been testing kiam in `us-gov-west-1` while using the `--region=us-gov-west-1` flag in the server DaemonSet manifest but I kept seeing this error:

`error retrieving credentials: InvalidClientTokenId: The security token included in the request is invalid`

when deploying an AWS CLI test pod to assume the pod role. Upon searching the closed issues here I came across this https://github.com/uswitch/kiam/issues/278 experiencing the same issue I was despite using the flag as intended.

Once I added `AWS_REGION` env var to the kiam server DaemonSet the pod was able to assume role successfully. 

It's my understanding that the `--region` option should handle any region-specific functions in kiam, is the requirement for the `AWS_REGION` env var intentional or could this be a bug?

darkdragn:
I'm thinking that this isn't the fault of their endpoint resolution. I'm thinking it's a side effect of the base AWS package for golang. I'm only bringing it up because I ran into the same issue for Vault's AWS Auth module, and even though I expressly set the region in the Auth config, it still didn't work unless I also set AWS_REGION.

I'm thinking that the endpoint resolution is needed for the DNS target, but AWS_REGION is being used for header signing on the requests or something to that effect.

pingles:
I've merged the above and will be released in v4. For now I'm going to close this and we can reopen if it hasn't fixed the issue.

PR #432, opened 2020-10-19
Improve region endpoint resolver

### LINKED ISSUE #410: kiam fails to resolve regional sts endpoint in us-iso-east-1 and us-isob-east-1
Hello!

We're deploying kiam in two of AWS's air gapped regions and we're finding that the sts endpoint fails to resolve.  

In us-iso-east-1, we're falling to this statement:  https://github.com/uswitch/kiam/blob/master/pkg/aws/sts/gateway.go#L73
And getting the error
```
Regional STS endpoint does not exist: sts.us-iso-east-1.amazonaws.com
```

After peeking around the code, I think we need special handling for these regions the same way china has special handling like here https://github.com/uswitch/kiam/blob/master/pkg/aws/sts/gateway.go

``` 
	if strings.HasPrefix(region, "cn-") {
		host = fmt.Sprintf("sts.%s.amazonaws.com.cn", region)
	} else if strings.HasPrefix(region, "us-isob") {
                host = fmt.Sprintf("sts.%s.sc2s.sgov.gov", region)
	} else if strings.HasPrefix(region, "us-iso") {
                host = fmt.Sprintf("sts.%s.c2s.ic.gov", region)
        } else {
		host = fmt.Sprintf("sts.%s.amazonaws.com", region)
	}
```

The Go AWS SDK handles these regions itself [(endpoints.json)](https://github.com/aws/aws-sdk-go-v2/blob/master/models/endpoints/endpoints.json#L8193-L8203) though so I wonder if there isn't some other issue?  I'm not familiar enough with the code to know for sure so sorry if this is some larger issue.  Thank you for any help you can provide!

pingles:
Pull out into separate type and add some more tests about handling of gov, fips and chinese regions.

Additionally removes comparison against SDK regions and instead relies on DNS resolution to verify.

This touches on, but doesn't necessarily fix #410 #368 #344 yet.


pingles:
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_enable-regions.html#id_credentials_temp_enable-regions_writing_code has more context on setting regions with STS

pkg/aws/sts/aws_endpoint_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+
+	if _, err := net.LookupHost(hostname); err != nil {
+		return "", fmt.Errorf("Regional STS endpoint does not exist: %s", hostname)
+	}
+
+	return hostname, nil
+}
+
+func newRegionalEndpointResolver(region string) (endpoints.Resolver, error) {
+	if region == "" || strings.Contains(region,"fips") {
```
Joseph-Irving, line 51: worth adding a comment about why this is the case

pkg/aws/sts/aws_endpoint_resolver.go
```diff
@@ -0,0 +1,73 @@
...
+func regionalHostname(region string) (string, error) {
+	hostname := fmt.Sprintf("sts.%s.amazonaws.com", region)
+
+	if strings.HasPrefix(region, "cn-") {
+		hostname = fmt.Sprintf("%s.cn", hostname)
+	}
+
+	if strings.HasPrefix(region, "us-iso") {
+		hostname = fmt.Sprintf("sts.%s.c2s.ic.gov", region)
+	}
```
Joseph-Irving, line 41: probably worth a comment explaining why the difference here

pkg/aws/sts/aws_endpoint_resolver.go
pingles: Added a note in the latest

pkg/aws/sts/aws_endpoint_resolver.go
pingles: Thanks, added a note in the latest commit

pingles:
Latest, this definitely fixes #410. 

It should also fix #368 which would have accessed the global endpoint while retrieving server credentials (and then using regional endpoints for pod credentials). The tests show it using a regional resolver and how it resolves the endpoints.

Joseph-Irving (approved):
lgtm

PR #435, opened 2020-10-27
Refactoring of policy and credentials

pingles:
* Rename CredentialsIdentity to RoleIdentity
* Remove arnResolver use within the credentials cache (it can work on the basis all credentials it's being used to manage have already got ARNs resolved)
* Policy interface changed to take role string and *Pod to avoid needing a PodGetter interface

@stefansedich hope you're ok with the rename of Credentials to Role? My reasoning was that the identity is largely used to refer to the role and not always when we have credentials.

### DIFF
--- a/pkg/aws/sts/arn_resolver.go
+++ b/pkg/aws/sts/arn_resolver.go
@@ -29,18 +29,24 @@ func DefaultResolver(prefix string) *Resolver {
 }
 
 // Resolve converts from a role string into the absolute role arn.
-func (r *Resolver) Resolve(role string) string {
+func (r *Resolver) Resolve(role string) (*RoleIdentity, error) {
 	if role == "" {
-		return ""
+		return nil, fmt.Errorf("role can't be empty")
+	}
+
+	if strings.HasPrefix(role, "arn:") {
+		return &RoleIdentity{ARN: role, Role: roleFromArn(role)}, nil
 	}
 
 	if strings.HasPrefix(role, "/") {
 		role = strings.TrimPrefix(role, "/")
 	}
 
-	if strings.HasPrefix(role, "arn:") {
-		return role
-	}
+	return &RoleIdentity{ARN: fmt.Sprintf("%s%s", r.prefix, role), Role: role}, nil
+}
 
-	return fmt.Sprintf("%s%s", r.prefix, role)
+// arn:aws:iam::account-id:role/role-name-with-path
+func roleFromArn(arn string) string {
+	splits := strings.SplitAfterN(arn, ":", 6)
+	return strings.TrimPrefix(splits[5], "role/")
 }
--- a/pkg/aws/sts/arn_resolver_test.go
+++ b/pkg/aws/sts/arn_resolver_test.go
@@ -17,55 +17,82 @@ import (
 	"testing"
 )
 
-func TestAddsPrefix(t *testing.T) {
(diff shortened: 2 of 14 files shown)


pingles:
This is also to help simplify things a little ahead of trying to incorporate #326 and #327.

stefansedich:
> * Rename CredentialsIdentity to RoleIdentity
> * Remove arnResolver use within the credentials cache (it can work on the basis all credentials it's being used to manage have already got ARNs resolved)
> * Policy interface changed to take role string and *Pod to avoid needing a PodGetter interface
> 
> @stefansedich hope you're ok with the rename of Credentials to Role? My reasoning was that the identity is largely used to refer to the role and not always when we have credentials.

I like it! `RoleIdentity` makes much more sense. Will this one be going in soon? I will rebase on this and then continue with adding the final tests to #430 so we can get that merged.

pingles:
Yeah hopefully. I keep making the refactorings and then changing my mind but should be in the next few days hopefully. Perhaps by the end of the week?

stefansedich:
> Yeah hopefully. I keep making the refactorings and then changing my mind but should be in the next few days hopefully. Perhaps by the end of the week?

Sounds good! I might hold off on getting to testing then incase more changes to avoid any merge pain, but once done will get the rest of my stuff done ASAP.

pingles:
I'll try and speed up then!

pingles:
I've done some refactoring and merged a previous PR #327 changes.

I want to write out a truth table to confirm and give examples but this will potentially be a breaking change. Because policy is now applied after the full ARN has been resolved the regexes will resolve differently (in both directions potentially). Regexes that denied pods previously may accidentally permit them, and regexes that previously denied may now permit. I suspect that the chances are low based on the way people would likely write them but before we merge I'd like to call out examples of where this behaviour would change.

I think this is still the right thing to do, but it will mean calling it out for cluster operators in the `docs/UPGRADING.md`.

@leosunmo what do you think? Could you prep some examples of how the namespace regex behaved before, and now, and what the implications are please?

I've not refactored to change policy so that it only uses resolved ARNs (rather than calling out to the resolver) but for now it's not essential.

stefansedich:
> @stefansedich this feels closer too; I changed code so that the `ARNResolver` converts from a role (either short-form or ARN) to a `RoleIdentity` type (which then has both `ARN` and `Role` fields).
> 
> What do you think? Next challenge is to add some more valuable tests to get codecov bot off my back :-) I've been meaning to write some better integration tests so may do that.

One thing that I am not sure off hand is in my PR for session-name and external-id I use `RoleIdentity` to add those additional bits of info, I think it should be OK I will likely just have to set those properties at a higher level after the call to `Resolve()`.

pingles:
Ah good point. Well I'll try and get it merged asap and then it's easier to integrate and see your work and ideas.

stefansedich:
> Ah good point. Well I'll try and get it merged asap and then it's easier to integrate and see your work and ideas.

Good stuff!

Joseph-Irving (approved):
I think this looks good, it would be good to understand what kind of breaking changes this may cause in relation to the regex so that we can call it out with big warning lights in the 4.0 changelog

pingles:
> it would be good to understand what kind of breaking changes this may cause in relation to the regex so that we can call it out with big warning lights in the 4.0 

Yep, totally agree! Will add an issue for it

PR #437, opened 2020-11-11
Replace -ms suffix from the gRPC keepalive flags to -duration

### LINKED ISSUE #354: gRPC keepalive flags imply the use of milliseconds
The flags used to configure the keepalive `time` and `timeout` imply the use of milliseconds rather than a duration.

```
  --grpc-keepalive-time-ms=10s   gRPC keepalive time
  --grpc-keepalive-timeout-ms=2s  
                                 gRPC keepalive timeout
```

Should probably have the `-ms` removed from the flag.

stefansedich:
Replace the `-ms` suffix from the keepalive flags to `-duration` remove confusion, this is a breaking change however as discussed with @pingles the 4.0 release includes other breaking changes so the time is right.

This will also require changes to the Helm chart, these will come post release to include any changes needed to the chart for 4.0 compatibility.

Closes #354

### DIFF
--- a/cmd/kiam/opts.go
+++ b/cmd/kiam/opts.go
@@ -96,8 +96,8 @@ type clientOptions struct {
 }
 
 func (o *clientOptions) bind(parser parser) {
-	parser.Flag("grpc-keepalive-time", "gRPC keepalive time").Default("10s").DurationVar(&o.keepaliveParams.Time)
-	parser.Flag("grpc-keepalive-timeout", "gRPC keepalive timeout").Default("2s").DurationVar(&o.keepaliveParams.Timeout)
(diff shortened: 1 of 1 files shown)


pingles:
I think we'd also need to update the Helm config and sample configs that reference these flags?

stefansedich:
> I think we'd also need to update the Helm config and sample configs that reference these flags?

I was not sure if helm changes here get deployed out as soon as this PR would be merged? I have opened an issue https://github.com/uswitch/kiam/issues/438 to track any required helm changes for a v4 release where they could be done at one time with a major chart version bump.

Happy to do with part of this PR but just myself was not sure if doing so would push an updated chart out now which could cause confusion.

pingles (approved):

Issue #438, opened 2020-11-11
Helm chart changes for v4 release


stefansedich:
Just tracking known changes for the chart for v4 release compatibility, these ideally are completed post release with a major bump for the chart version.

- [x] Replace -ms suffix from the gRPC keepalive flags to -duration https://github.com/uswitch/kiam/pull/437
- [x] Add support for configuring gRPC max-connection-age for the kiam server https://github.com/uswitch/kiam/pull/440

PR #439, opened 2020-11-12
Helm chart v4 release compatibility

stefansedich:
This PR makes the changes required to the chart for full compatiblity with the v4 release, to be merged at or post ther v4 release.

Closes https://github.com/uswitch/kiam/issues/438

### DIFF
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: kiam
-version: 5.10.0
-appVersion: 3.6
+version: 6.0.0
+appVersion: 4.0
 description: Integrate AWS IAM with Kubernetes
 keywords:
   - kiam
--- a/helm/kiam/README.md
+++ b/helm/kiam/README.md
@@ -163,7 +163,7 @@ The following table lists the configurable parameters of the kiam chart and thei
 | `agent.image.tag`                           | Agent image tag                                                                              | `v3.6`                       |
 | `agent.image.pullPolicy`                    | Agent image pull policy                                                                      | `IfNotPresent`               |
 | `agent.dnsPolicy`                           | Agent pod DNS policy                                                                         | `ClusterFirstWithHostNet`    |
-| `agent.whiteListRouteRegexp`                | Agent pod whitelist metadata API path argument regex                                         | `{}`                         |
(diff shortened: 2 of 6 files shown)


stefansedich:
@pingles will keep this one in draft until ready, I was thinking all v4 related changes could just be rolled into this single PR if you think that works?

pingles:
Do we want to push a version of this with the 4.0-beta2 image?

Or better to just wait to push v4.0 and then this, @Joseph-Irving any thoughts also? (I'm not super familiar with Helm workflows).

pingles:
Am I right in thinking that this means the current Helm chart is broken (given #443)

stefansedich:
> Am I right in thinking that this means the current Helm chart is broken (given #443)

So that looks to be the case pulling the 5.10.0 chart and looking at the contents, my theory here is:

1. #427 was merged making chart changes and not bumping the version
2. CI built and pushed the 5.10.0 chart again with the now breaking changes
3. :cry: 

I would recommend rolling back the chart changes without a version bump so we get a good 5.10.0 pushed again and then rolling the changes into my since PR that adds v4 compatibility with a major version bump.

I am not overly familiar with the CI process but that is what it looks like to me @pingles

Joseph-Irving:
yeah I agree, I think the easiest way to fix the chart is to revert that PR https://github.com/uswitch/kiam/pull/427, which will cause 5.10 to be recreated in a fixed state. We should definitely get that PR back into the helm 6.0 release as it's a breaking change.
I'd probably just wait until we have the 4.0 image @pingles, don't think it's normal to have pre-release images in helm charts generally

pingles:
Thanks @Joseph-Irving. I'd rather not revert that PR if we can avoid it. Can we update the build/Helm charts to fix it also?

pingles:
Or do you mean we just revert the Helm changes that were made as part of that PR?

Joseph-Irving:
yeah just reverting the helm section should fix it, we should make a new PR or incorporate it into this one to reimplement those changes with a version bump

stefansedich:
> yeah just reverting the helm section should fix it, we should make a new PR or incorporate it into this one to reimplement those changes with a version bump

Done I incorporated the `--allow-route-regexp` changes into this PR.

pingles:
Thanks @stefansedich. I've marked as the v4 milestone so should be easier to see what remains now.

pingles (approved):

Issue #433, opened 2020-10-22
Allow MaxConnectionAge on Server


pingles:
gRPC issues highlighted in https://github.com/uswitch/kiam/issues/425 mention a preference to controlling flow of updates of the connection addresses via `MAX_CONNECTION_AGE`.

This should be added as a flag to the server, and passed when creating the gRPC server:

```
grpc.NewServer(grpc.KeepaliveParams(keepalive.ServerParameters{MaxConnectionAge: maxConnAge}))
```

The default value is infinity (https://github.com/grpc/grpc-go/blob/master/keepalive/keepalive.go#L60) but could be overridden to force clients to re-resolve and reconnect more regularly.

Should help #425 along with updating the gRPC Go version.

stefansedich:
@pingles do we want to keep the existing default of `infinity` or should we default to something more sensible that suits our needs better?

PR #440, opened 2020-11-12
Add support for configuring gRPC max-connection-age for the kiam server

stefansedich:
This PR adds support for configuring the `max-connection` related options for the kiam server.

Closes https://github.com/uswitch/kiam/issues/433

### DIFF
--- a/cmd/kiam/server.go
+++ b/cmd/kiam/server.go
@@ -15,6 +15,8 @@ package main
 
 import (
 	"context"
+	"fmt"
+	"math"
 	"os"
 	"os/signal"
 	"syscall"
@@ -46,6 +48,8 @@ type serverOptions struct {
 }
 
 func (o *serverOptions) bind(parser parser) {
+	infiniteDuration := fmt.Sprintf("%ds", math.MaxInt64/1000000000)
+
 	parser.Flag("fetchers", "Number of parallel fetcher go routines").Default("8").IntVar(&o.ParallelFetcherProcesses)
 	parser.Flag("prefetch-buffer-size", "How many Pod events to hold in memory between the Pod watcher and Prefetch manager.").Default("1000").IntVar(&o.PrefetchBufferSize)
 	parser.Flag("bind", "gRPC bind address").Default("localhost:9610").StringVar(&o.BindAddress)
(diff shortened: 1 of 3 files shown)


stefansedich:
@pingles curios on your thoughts here, I added support for all of the `max-connection` options, the one thing I could not see how to do cleaner was to default to infinity for the values.

pingles:
Thanks. Not sure how to better handle defaults, although perhaps the infinity defaults would be better to be limited anyway?

The failure mode occurs when a server fails to resolve addresses (or server addresses are removed) but the load-balance won't re-fill the pool unless a connection change occurs. So the gRPC team's advice was to set those settings to ensure it triggers a reconnection.

I wanted to get some tests to try and prove the failing behaviour today, and managed to get something rolling with https://github.com/uswitch/kiam/pull/441.

stefansedich:
> Thanks. Not sure how to better handle defaults, although perhaps the infinity defaults would be better to be limited anyway?
> 
> The failure mode occurs when a server fails to resolve addresses (or server addresses are removed) but the load-balance won't re-fill the pool unless a connection change occurs. So the gRPC team's advice was to set those settings to ensure it triggers a reconnection.
> 
> I wanted to get some tests to try and prove the failing behaviour today, and managed to get something rolling with #441.

I was not sure of a good default here, but you are right we can do something more reasonable for "infinity" I am sure 1 month is even enough here? or we could pick a reasonable default that we know works whether that be 1 minute or something else we decide on.

pingles:
I would've thought something on the order of 15 minutes is a good starting point. https://github.com/pomerium/pomerium set it to 5 minutes.

stefansedich:
> I would've thought something on the order of 15 minutes is a good starting point. https://github.com/pomerium/pomerium set it to 5 minutes.

Done!

cmd/kiam/server.go
```diff
@@ -59,6 +59,11 @@ func (o *serverOptions) bind(parser parser) {
 	parser.Flag("session-refresh", "How soon STS Tokens should be refreshed before their expiration.").Default("5m").DurationVar(&o.SessionRefresh)
 	parser.Flag("assume-role-arn", "IAM Role to assume before processing requests").Default("").StringVar(&o.AssumeRoleArn)
 	parser.Flag("region", "AWS Region to use for regional STS calls (e.g. us-west-2). Defaults to the global endpoint.").Default("").StringVar(&o.Region)
+	parser.Flag("grpc-keepalive-time-duration", "gRPC keepalive time").Default("10s").DurationVar(&o.KeepaliveParams.Time)
```
stefansedich, line 62: I took the defaults for the time/timeout from the existing agent options.

cmd/kiam/server.go
```diff
@@ -59,6 +59,11 @@ func (o *serverOptions) bind(parser parser) {
...
+	parser.Flag("grpc-keepalive-timeout-duration", "gRPC keepalive timeout").Default("2s").DurationVar(&o.KeepaliveParams.Timeout)
+	parser.Flag("grpc-max-connection-idle-duration", "gRPC max connection idle").Default("15m").DurationVar(&o.KeepaliveParams.MaxConnectionIdle)
```
stefansedich, line 64: Defaulted max-connection options to 15m as discussed.

pingles:
Thanks @stefansedich. I think some of those defaults may need to increase but let me pull the code and check.

pingles:
Actually, looks ok to me. I think it'd be good to get some real-world usage and see how it behaves but seems sensible to me to start with these.

Thanks again.

pingles (approved):

Issue #425, opened 2020-09-16
Agent stuck  with log  "resolver returned no addresses"


avdhoot:
Some of the kiam-agent does not return metadata & emit below logs. We also monitored network activity of agent it never makes new DNS query.

Agent logs:
```{"level":"warning","msg":"error finding role for pod: rpc error: code = Unavailable desc = resolver returned no addresses","pod.ip":"100.121.119.112","time":"2020-09-16T09:51:07Z"}```

To fix this we delete kiam-agent pod.

**Comment by BartoszZawadzki** [comment]
We are facing the same issue as above:

`kiam-agent-2vb9x kiam-agent {"addr":"100.96.1.231:44236","level":"error","method":"GET","msg":"error processing request: rpc error: code = Unavailable desc = resolver returned no addresses","path":"/latest/meta-data/iam/security-credentials/","status":500,"time":"2020-10-14T12:48:06Z"}`

kiam v3.6
kubernetesVersion: 1.18.9
kops 1.18.1 (git-453d7d96be)

And yes, killing agent pods seemed to resolve the issue (for now)

**Comment by pingles** [comment]
The error means that the agent pod's gRPC client is unable to find any available server pods for it to connect to. It uses gRPC's client load-balancer. There's some gRPC environment variables you can add for it to report debugging information as to what's going on but the problem is that there's no running server pods and it wasn't able to find them in time.

**Comment by dza89** [comment]
We are facing this issue as well. All kiam agents became unresponsive until a restart.
kiam v3.6
kubernetesVersion: 1.17.eks.3

@pingles 
That makes sense, but shouldn't it lead to an application exit if it's 'stuck'?

**Comment by pingles** [comment]
It depends where it happens. Most operations are set to retry in the event of failures (assuming some intermittent pod issues or DNS resolution for example) up to an amount of time or the operation is cancelled. 

Other issues have highlighted similar problems and you can enable some env vars to increase the verbosity of gRPC's logging. https://github.com/uswitch/kiam/issues/217#issuecomment-521933917

Could you try that and see what that shows please.

**Comment by dagvl** [comment]
I am also observing this problem. Here are some detailed logs:

Logs from kiam-agent before problem:
```
➜ k logs -f kiam-agent-5n76g
{"level":"info","msg":"configuring iptables","time":"2020-10-20T21:58:12Z"}
{"level":"info","msg":"started prometheus metric listener 0.0.0.0:9620","time":"2020-10-20T21:58:12Z"}
[...]

PR #441, opened 2020-11-12
gRPC Resolution

pingles:
This is to help build some tests around the behaviour of gRPC's load balancer and better manage recovery when server addresses don't resolve.


pingles:
Forgot to add, this is to try and help address issues highlighted in  #217 and #425

PR #447, opened 2020-12-07
Log session-name and external-id when logging credentials

stefansedich:
Adds the session-name and external-id to the log fields when logging credentials.

Session name is safe and easy but logging external-id worries me a little, at the moment I mask the entire string which really adds no value apart from showing me there is an external-id used and how long it might be.

We could attempt some partial masking but that again makes me nervous, thoughts @pingles ?

### DIFF
--- a/pkg/aws/sts/log.go
+++ b/pkg/aws/sts/log.go
@@ -14,13 +14,25 @@
 package sts
 
 import (
+	"strings"
+
 	log "github.com/sirupsen/logrus"
 )
 
 func CredentialsFields(identity *RoleIdentity, creds *Credentials) log.Fields {
-	return log.Fields{
+	fields := log.Fields{
 		"credentials.access.key": creds.AccessKeyId,
 		"credentials.expiration": creds.Expiration,
 		"credentials.role":       identity.Role,
 	}
+
+	if identity.SessionName != "" {
+		fields["credentials.session-name"] = identity.SessionName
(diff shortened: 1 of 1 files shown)


pkg/aws/sts/log.go
```diff
@@ -14,13 +14,25 @@
...
 	log "github.com/sirupsen/logrus"
 )
 
 func CredentialsFields(identity *RoleIdentity, creds *Credentials) log.Fields {
-	return log.Fields{
+	fields := log.Fields{
 		"credentials.access.key": creds.AccessKeyId,
 		"credentials.expiration": creds.Expiration,
-		"credentials.role":       identity.Role,
+		"credentials.role":       identity.Role.ARN,
```
stefansedich, line 26: We were logging the entire Role object which isn't nescasary, this will log only the ARN string and not the nested object.

pingles:
I think logging the masked version is ok. 

As you say, it confirms a value is set. As per the AWS docs (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html), it's not considered a secret as the org is expected to issue them. But, for it not to be a problem I'd expect logs to need to be secured properly and make sure people couldn't see all IDs together.

For now, I'd say let's keep the masked version and then if it's problematic we can address later.

pingles (approved):

Issue #443, opened 2020-12-01
Why not upgrade the helm chart version after changed the templates?


wd:
This a diff from same helm chart version `5.10.0`, the `kiam.orig` was downloaded 3 months ago. This is break change, I think it often need a major version or at least a minor version bump. 

```
diff -ruN kiam.orig/README.md new/kiam/README.md
--- kiam.orig/README.md 2020-08-18 23:43:42.000000000 +0800
+++ new/kiam/README.md  2020-10-15 19:44:45.000000000 +0800
@@ -163,7 +163,7 @@
 | `agent.image.tag`                           | Agent image tag                                                                              | `v3.6`                       |
 | `agent.image.pullPolicy`                    | Agent image pull policy                                                                      | `IfNotPresent`               |
 | `agent.dnsPolicy`                           | Agent pod DNS policy                                                                         | `ClusterFirstWithHostNet`    |
-| `agent.whiteListRouteRegexp`                | Agent pod whitelist metadata API path argument regex                                         | `{}`                         |
+| `agent.allowRouteRegexp`                    | Agent metadata proxy server only allows accesses to paths matching this regexp               | `{}`                         |
 | `agent.extraArgs`                           | Additional agent container arguments                                                         | `{}`                         |
 | `agent.extraEnv`                            | Additional agent container environment variables                                             | `{}`                         |
 | `agent.extraHostPathMounts`                 | Additional agent container hostPath mounts                                                   | `[]`                         |
diff -ruN kiam.orig/templates/agent-daemonset.yaml new/kiam/templates/agent-daemonset.yaml
--- kiam.orig/templates/agent-daemonset.yaml    2020-08-18 23:43:42.000000000 +0800
+++ new/kiam/templates/agent-daemonset.yaml     2020-10-15 19:44:45.000000000 +0800
@@ -106,8 +106,8 @@
             - --prometheus-listen-addr=0.0.0.0:{{ .Values.agent.prometheus.port }}
             - --prometheus-sync-interval={{ .Values.agent.prometheus.syncInterval }}
             {{- end }}
-            {{- if .Values.agent.whiteListRouteRegexp }}
-            - --whitelist-route-regexp={{ .Values.agent.whiteListRouteRegexp }}
+            {{- if .Values.agent.allowRouteRegexp }}
+            - --allow-route-regexp={{ .Values.agent.allowRouteRegexp }}
             {{- end }}
[...]

PR #448, opened 2020-12-10
Revert helm breakage from #427

pingles:
From conversation in #439, this should fix the Helm chart breakage introduced in #427

Issue was also reported in #443.


pingles:
@Joseph-Irving @stefansedich hopefully this fixes it.

Joseph-Irving (approved):

stefansedich:
@pingles can confirm the 5.10.0 chart looks fixed now, I also incorporated these changed into https://github.com/uswitch/kiam/pull/439 so we should be good for v4 compat once that is merged.

PR #455, opened 2021-01-05
fixing two typos in readme

frodopwns:
Couple of things I noticed during my reading of the docs:

maybe -> may be

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ metadata:
 ```
 
 You can also control the external id used when assuming the role via an annotation added to the `Pod`, which
-maybe used to avoid [confused deputy scenarios in cross-organisation role assumption](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html). For example:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,7 +61,7 @@ metadata:
 ```
 
 You can also control the external id used when assuming the role via an annotation added to the `Pod`, which
-maybe used to avoid [confused deputy scenarios in cross-organisation role assumption](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html). For example:
+may be used to avoid [confused deputy scenarios in cross-organization role assumption](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html). For example:
```
Joseph-Irving, line 64: organisation is the correct spelling in UK English

Joseph-Irving: haha, it's ok

Joseph-Irving (approved):
lgtm, thanks!

Issue #456, opened 2021-01-05
Update Helm chart repo


pingles:
The build is currently broken because it can't push Helm charts.

```
+ sh ci/helm-package.sh
fetch http://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/APKINDEX.tar.gz
fetch http://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
(1/6) Installing brotli-libs (1.0.9-r2)
(2/6) Installing nghttp2-libs (1.42.0-r0)
(3/6) Installing libcurl (7.74.0-r0)
(4/6) Installing expat (2.2.10-r0)
(5/6) Installing pcre2 (10.36-r0)
(6/6) Installing git (2.30.0-r0)
Executing busybox-1.30.1-r2.trigger
OK: 19 MiB in 21 packages
diff --git a/helm/kiam/Chart.yaml b/helm/kiam/Chart.yaml
--- a/helm/kiam/Chart.yaml
diff --git a/helm/kiam/README.md b/helm/kiam/README.md
--- a/helm/kiam/README.md
diff --git a/helm/kiam/templates/agent-daemonset.yaml b/helm/kiam/templates/agent-daemonset.yaml
--- a/helm/kiam/templates/agent-daemonset.yaml
diff --git a/helm/kiam/templates/server-daemonset.yaml b/helm/kiam/templates/server-daemonset.yaml
--- a/helm/kiam/templates/server-daemonset.yaml
diff --git a/helm/kiam/templates/server-deployment.yaml b/helm/kiam/templates/server-deployment.yaml
--- a/helm/kiam/templates/server-deployment.yaml
diff --git a/helm/kiam/values.yaml b/helm/kiam/values.yaml
--- a/helm/kiam/values.yaml
Creating /root/.helm 
Creating /root/.helm/repository 
Creating /root/.helm/repository/cache 
Creating /root/.helm/repository/local 
Creating /root/.helm/plugins 
Creating /root/.helm/starters 
Creating /root/.helm/cache/archive 
Creating /root/.helm/repository/repositories.yaml 
Adding stable repo with URL: https://kubernetes-charts.storage.googleapis.com 
Error: error initializing: Looks like "https://kubernetes-charts.storage.googleapis.com" is not a valid chart repository or cannot be reached: Failed to fetch https://kubernetes-charts.storage.googleapis.com/index.yaml : 403 Forbidden
```

[...]

PR #457, opened 2021-01-05
change helm init repo to charts.helm.sh

Joseph-Irving:
Seems the helm repo was switched from https://kubernetes-charts.storage.googleapis.com to https://charts.helm.sh/stable at some point, this should hopefully fix the build 

Fixes: https://github.com/uswitch/kiam/issues/456

### DIFF
--- a/ci/helm-package.sh
+++ b/ci/helm-package.sh
@@ -12,7 +12,7 @@ then
   exit 0
 else
   # package helm chart
-  helm init --client-only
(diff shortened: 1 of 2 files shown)


Joseph-Irving:
We'll need to cause some kind of change to the helm file too, otherwise CI won't try to republish

Joseph-Irving:
I've update the helm changelog, that should cause it to republish

pingles:
Thanks @Joseph-Irving!

pingles (approved):

PR #474, opened 2021-05-07
Added helm chart value option to disable strict namespace regexp mode introduced in V4

wcarlsen:
I think the introduction of strict namespace regexp mode is a great addition to KIAM, hence it should be kept default. But configuring the flag should be more straight forward.

Thanks for the great work.

### DIFF
--- a/helm/kiam/Chart.yaml
+++ b/helm/kiam/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 6.0.0
+version: 6.0.1
 appVersion: 4.0
 description: Integrate AWS IAM with Kubernetes
 keywords:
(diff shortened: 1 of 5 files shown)


avnes (dismissed):

helm/kiam/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kiam
-version: 6.0.0
+version: 6.0.1
```
Joseph-Irving, line 3: can you bump to 6.1 as this is not a bug fix

wcarlsen: Yeah sure. Good point. This should be fixed now.

Joseph-Irving (approved):
lgtm, thanks

PR #487, opened 2021-07-14
Add agent.dnsConfig to Helm chart values

KevinSnyderCodes:
In some GKE clusters, there is a bug where `hostNetwork: true` and `dnsPolicy: ClusterFirstWithHostNet` results in failed DNS queries.

The solution is to set `dnsPolicy: None` and provide your own `dnsConfig`.

This PR adds `agent.dnsConfig` to the Helm chart values, allowing users to implement this workaround to the GKE issue or specify any other `dnsConfig` they wish.

https://cloud.google.com/kubernetes-engine/docs/how-to/nodelocal-dns-cache#dns_timeout


KevinSnyderCodes:
Closed as AWS and GCP both use `169.254.169.254` for their Metadata API, making kops incompatible with GKE Workload Identity.

We ended up using [gtoken](https://github.com/doitintl/gtoken) as an alternative.

PR #501, opened 2021-09-21
Use v1 RBAC objects in helm chart

### LINKED ISSUE #500: Helm chart uses deprecated v1beta1 RBAC objects
The helm chart uses the deprecated `rbac.authorization.k8s.io/v1beta1` api for `ClusterRole` and `ClusterRoleBinding` objects. This causes warnings while installing the helm chart, and means that it will not work with v1.22+ Kubernetes clusters.

fhke:
resolve #500


Joseph-Irving:
Can you bump the chart version please https://github.com/uswitch/kiam/blob/master/helm/kiam/Chart.yaml

PR #507, opened 2021-10-26
upgrade to v1.20 libraries, fixes issue#484

### LINKED ISSUE #484: Events missing on pods due to use of deprecated metadata.selfLink
Since upgrading to Kubernetes `1.20` we no longer get `kiam` related events on pods due to the use of the deprecated `metadata.selfLink` field. This was deprecated in Kubernetes `1.16` and [removed](https://github.com/kubernetes/kubernetes/pull/94397) in `1.20`

We see error logs relating to this in `kiam-server`, example:

```
ERROR: logging before flag.Parse: E0706 15:08:28.333117       1 event.go:260] Could not construct reference to: '&v1.Pod{TypeMeta:v1.TypeMeta{Kind:"", APIVersion:""}, 
<redacted>
 due to: 'selfLink was empty, can't make reference'. Will not report event: 'Warning' 'KiamCredentialError' 'failed retrieving credentials: AccessDenied: User: arn:aws:sts::<redacted>:assumed-role/<redacted>/<redacted> is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::<redacted>:role/<redacted>'
```

Looks like it's using `EventRecorder` from `client-go` although `v7.0.0` looks pretty old, I'm guessing the client libraries will need updating to use a version that constructs the object reference using different fields.

jjo:
Fixes #484.

* upgrade to k8s.io/client-go@v0.20.0 (and deps)
* tweak Dockerfile to use generated (and committed) proto/service.pb.go,
  rather than doing it at build time -- as *it's already committed*,
  it makes built artifacts deterministic/reproduceable from src tree
* use `golang:1.15.15`, same as upstream Kubernetes v1.20


Dockerfile
```diff
@@ -14,6 +14,8 @@ COPY pkg/ pkg/
 COPY proto/ proto/
 COPY Makefile Makefile
 
+# As (generated) proto/service.pb.go is _also_ committed, void the need to install protoc / protoc-gen-go plugin
+RUN touch proto/service.pb.go
```
Joseph-Irving, line 18: can you explain this change a bit more? I don't quite follow, we already copy the whole proto directory into the build container as per line 14

jjo: sure, but depending on the `mtime` timestamp from the copy, it _may_ trigger rebuilding `proto/service.pb.go` which _may_ endup producing a different file from the corresponding `Makefile` target (it _shouldn't_, but ...) -- ie I want to be 100% sure we're using the committed file, also not enter into the biz of `go get ...`ing the protoc compiler at docker build time.

Joseph-Irving (approved):
lgtm, thanks

